Issue #23, opened 2025-06-07
5 hours sessions tracker


ryoppippi:
### Discussed in https://github.com/ryoppippi/ccusage/discussions/20

<div type='discussions-op-text'>

<sup>Originally posted by **IntegralMedia01** June  7, 2025</sup>
please add ability to keep track of those 5 hours sessions. how many sessions already used for the month.</div>

a-c-m:
I'm interested in this. 

Would like to be able to see:
- Session start/end
- tokens used, by model.
- % of allocation used (if it can tell the tier you are on, or is provided by cli)

If a session is active, it should also predict, total session usage (i.e. will you hit the limit) for that session, by simple extrapolation is fine.

ryoppippi:
Thank you everyone.
Actually this feature is once closed by https://github.com/ryoppippi/ccusage/pull/55 , so I need to figure out which one is better.
I'll review both of them  and decide one of your works. 
Thank you for your contributions! Happy vibe coding!

Kaniikura:
Hi @ryoppippi and @a-c-m,

My apologies for the duplicate PR. After reviewing the technical specifications more carefully, I realize that PR #80's approach with dynamic 5-hour windows starting from first message is the correct implementation of Claude's session tracking.

My PR #84 incorrectly uses fixed UTC time windows, which doesn't match how Claude actually counts sessions. @a-c-m's implementation properly follows the rolling 5-hour window specification.

I'll close my PR since #80 has the correct approach. Great work @a-c-m\!

Thanks for your patience with the confusion.

PR #55, opened 2025-06-11
5 hour session window usage info

EphemeralSapient:
fixes: #20 
fixes: #23 
Used claude code to implement that 5 hour session window thing, let me know if its okay

## Usage Examples

  ```bash
  # Show session usage and limits
  ccusage sessions

  # Filter by date range
  ccusage sessions --since 20250101 --until 20250131

  # Export as JSON for programmatic access
  ccusage sessions --json

  # Get session count programmatically
  ccusage sessions --json | jq '.sessions[0].totalSessions'
```

## Summary by CodeRabbit

- **New Features**
  - Introduced a "Session Windows Report" for the Claude Max plan, allowing users to view and track usage in 5-hour session windows with a monthly session limit of 50 sessions.
  - Added a new CLI command to generate session-based usage reports, including detailed metrics, color-coded alerts, and summary statistics.
  - Enhanced reporting options with date filtering, JSON output, cost calculation modes, and customizable sort order.

- **Documentation**
  - Updated the README to include instructions and examples for the new session windows report, CLI usage, and key features.

- **Tests**
  - Added comprehensive tests to validate session window grouping, aggregation, and edge cases.

- **Chores**
  - Reformatted package configuration for consistency (no functional changes).

- **Style**
  - Changed default sort order for reports to show newest data first.

### CODE REVIEW

File: src/commands/sessions.ts [hunk 1]
```diff
@@ -0,0 +1,206 @@
...
+				// Month header with utilization stats
+				const utilizationColor = monthStats.utilizationPercent > 80
+					? 'red'
+					: monthStats.utilizationPercent > 60
+						? 'yellow'
+						: 'green';
+
+				logger.info(`\n${pc.bold(monthStats.month)} - ${pc[utilizationColor](`${monthStats.totalSessions} sessions used`)} (${pc[utilizationColor](`${monthStats.utilizationPercent.toFixed(1)}%`)} of 50 limit)`);
+
+				if (monthStats.remainingSessions <= 5 && monthStats.remainingSessions > 0) {
```
ryoppippi: what does `5` mean?

ryoppippi:
@EphemeralSapient Hi
Thank you for this PR. Awesome.

I found that there is a magic number `remainingSessions`. 50 session limit is only for $100 plan and $200 plan has different limitation, right? If I'm wrong explain it.

src/types.internal.ts
```diff
@@ -11,3 +11,45 @@ export type CostMode = TupleToUnion<typeof CostModes>;
...
+	utilizationPercent: v.number(), // (totalSessions / 50) * 100
+	totalCost: v.number(),
+	totalTokens: v.number(),
+	averageCostPerSession: v.number(),
+	averageTokensPerSession: v.number(),
+	windows: v.array(SessionWindowSchema),
+	currentSession: v.optional(CurrentSessionInfoSchema),
+});
+
+export type SessionWindowStats = v.InferOutput<typeof SessionWindowStatsSchema>;
```
ryoppippi: you can define those types in `data-loader.ts` because we don't use it in other files.
also you can stop exporting it

EphemeralSapient:
ah, forgot about that completely, I thought both 100$ and 200$ plan had same session; I was wrong

followed up with above review; again used CC , sorry if I missed anything :P

ryoppippi:
@EphemeralSapient I came up an idea.
Instead of creating new sub-command, how about add `--current` option to `session` sub-command?

EphemeralSapient:
after using it a while; I found inconsistent values on window start timing, I didn't notice it earlier

Honestly I'm not sure what to do here 😖

ryoppippi:
> after using it a while; I found inconsistent values on window start timing, I didn't notice it earlier

OMG 

Thank you for investigating.

Well if you cannot make sure that it is correct, we gonna suspend this PR. but thank you for you work

EphemeralSapient:
sorry for the mess, I'll be closing this PR

Have a nice day

PR #58, opened 2025-06-11
fix: implement chronological deduplication for branched conversations

### LINKED ISSUE #19: [Bug] Branched conversations are fully counted
When you branch a conversation in CC, when pressing Esc twice for example, a new JSONL gets created that is very similar to the initial one.
However, obviously it doesn't have the same cost associated with it (previous output tokens are becoming input tokens for example). The combination of message id + request id of assistant message should only be counted once globally, as it heavily skews the numbers otherwise.
You can test that behavior by branching a very long, output intensive conversation multiple times and calculating the price, it's way more than it should be. Only the delta (new assistant message that contains the previous output as input) in the new conversations should be included.

ben-vargas:
# Fix: Implement Chronological Deduplication for Branched Conversations

## Summary

This PR fixes issue #19 by implementing a chronologically-aware deduplication system that prevents double-counting of tokens and costs when conversations are branched in Claude Code.

Fixes #19

## Problem

When users branch conversations in Claude Code (by editing a past message with Esc+Esc), the application creates a new conversation that includes duplicate entries from the original conversation. These duplicates have identical `requestId` and `message.id` values, causing:

- **Inflated token counts** - The same API calls are counted multiple times
- **Incorrect cost calculations** - Users see higher costs than actually incurred
- **Wrong date attribution** - Costs may be attributed to the branch date instead of the original date

## Root Cause

The current implementation processes JSONL files in an undefined order (filesystem-dependent). This means:

1. A newer branched conversation file might be processed before the original
2. Without deduplication, both files contribute to the totals
3. Even with simple deduplication, processing order matters - the first processed entry "wins"

## Solution

This PR implements a comprehensive deduplication solution that:

### 1. **Chronological File Processing**
- Added `getEarliestTimestamp()` to extract the earliest timestamp from each JSONL file
- Added `sortFilesByTimestamp()` to sort files by their earliest timestamp
- Files are now processed oldest-first, ensuring original entries are preserved

### 2. **Message Deduplication**
- Extended `UsageDataSchema` to include `message.id` and `requestId` fields
- Added `createUniqueHash()` to generate unique identifiers from message+request IDs
- Track processed hashes in a Set to skip duplicate entries
- Implemented in both `loadDailyUsageData()` and `loadSessionData()`

### 3. **Robust Implementation**
- Handles files without timestamps (placed at end of processing order)
- Skips invalid JSON lines gracefully
- Maintains backward compatibility with existing data

## Testing

Added comprehensive test suite (`data-loader-deduplication.test.ts`) with 11 tests covering:

- ✅ Hash creation from message and request IDs
- ✅ Timestamp extraction from JSONL files
- ✅ File sorting by timestamp
- ✅ Deduplication in daily usage data
- ✅ Deduplication in session data
- ✅ Edge cases (missing IDs, invalid JSON, no timestamps)

All existing tests (117 total) continue to pass.

### Modified Files:
- `src/data-loader.ts` - Added deduplication logic and chronological processing
- `src/data-loader-deduplication.test.ts` - New comprehensive test suite

### Key Functions Added:
```typescript
// Create unique identifier for deduplication
createUniqueHash(data: UsageData): string | null

// Extract earliest timestamp from JSONL file
getEarliestTimestamp(filePath: string): Promise<Date | null>

// Sort files chronologically
sortFilesByTimestamp(files: string[]): Promise<string[]>
```

## Impact

- **Accurate Costs**: Users will see correct token counts and costs
- **Proper Attribution**: Usage is attributed to the original date, not branch date
- **No Data Loss**: Only true duplicates are removed
- **Performance**: Minimal overhead from sorting files before processing

## Verification

Before this fix:
- Branched conversations would show inflated costs
- Same usage could appear on multiple dates

After this fix:
- Each unique API call is counted exactly once
- Usage appears only on the original date
- Branched conversations no longer inflate costs

## Breaking Changes

None. This change is backward compatible and transparent to users.

## Future Considerations

This implementation uses a simple concatenation for the hash (`messageId:requestId`). If needed, this could be upgraded to use a proper hashing algorithm, though the current approach is sufficient for the use case.

### DIFF
--- a/src/data-loader-deduplication.test.ts
+++ b/src/data-loader-deduplication.test.ts
@@ -1,0 +1,334 @@
+import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+import { describe, expect, it } from 'bun:test';
+import {
+	createUniqueHash,
+	getEarliestTimestamp,
+	loadDailyUsageData,
+	loadSessionData,
+	sortFilesByTimestamp,
+} from './data-loader.ts';
+
+describe('deduplication functionality', () => {
+	describe('createUniqueHash', () => {
+		it('should create hash from message id and request id', () => {
+			const data = {
+				timestamp: '2025-01-10T10:00:00Z',
+				message: {
+					id: 'msg_123',
+					usage: {
+						input_tokens: 100,
+						output_tokens: 50,
+					},
+				},
+				requestId: 'req_456',
+			};
+
+			const hash = createUniqueHash(data);
+			expect(hash).toBe('msg_123:req_456');
+		});
+
+		it('should return null when message id is missing', () => {
+			const data = {
+				timestamp: '2025-01-10T10:00:00Z',
+				message: {
+					usage: {
+						input_tokens: 100,
+						output_tokens: 50,
+					},
+				},
+				requestId: 'req_456',
+			};
+
+			const hash = createUniqueHash(data);
+			expect(hash).toBeNull();
+		});
+
+		it('should return null when request id is missing', () => {
+			const data = {
+				timestamp: '2025-01-10T10:00:00Z',
+				message: {
+					id: 'msg_123',
+					usage: {
+						input_tokens: 100,
+						output_tokens: 50,
+					},
+				},
+			};
+
+			const hash = createUniqueHash(data);
+			expect(hash).toBeNull();
+		});
+	});
+
+	describe('getEarliestTimestamp', () => {
+		let tempDir: string;
+
+		it('should extract earliest timestamp from JSONL file', async () => {
+			tempDir = await mkdtemp(path.join(tmpdir(), 'ccusage-test-'));
+			const testFile = path.join(tempDir, 'test.jsonl');
+
+			const content = [
+				JSON.stringify({ timestamp: '2025-01-15T12:00:00Z', message: { usage: {} } }),
+				JSON.stringify({ timestamp: '2025-01-10T10:00:00Z', message: { usage: {} } }),
+				JSON.stringify({ timestamp: '2025-01-12T11:00:00Z', message: { usage: {} } }),
+			].join('\n');
+
+			await writeFile(testFile, content);
+
+			const timestamp = await getEarliestTimestamp(testFile);
+			expect(timestamp).toEqual(new Date('2025-01-10T10:00:00Z'));
+
+			await rm(tempDir, { recursive: true });
+		});
+
+		it('should handle files without timestamps', async () => {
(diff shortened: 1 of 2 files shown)


src/data-loader.ts
```diff
@@ -99,6 +101,91 @@ export function formatDate(dateStr: string): string {
...
+				// Skip invalid JSON lines
+				continue;
+			}
+		}
+
+		return earliestDate;
+	}
+	catch {
+		return null;
+	}
```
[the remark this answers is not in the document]
ben-vargas: Improved as recommended - commit c8d206b

ben-vargas:
Curious to see how this runs for others... here's my macbook pro:

New Output:
CleanShot 2025-06-11 at 13 28 05@2x

Old Output:
CleanShot 2025-06-11 at 13 28 55@2x

src/data-loader-deduplication.test.ts
```diff
@@ -0,0 +1,334 @@
...
+				'{ broken: json',
+			].join('\n');
+
+			await writeFile(testFile, content);
+
+			const timestamp = await getEarliestTimestamp(testFile);
+			expect(timestamp).toEqual(new Date('2025-01-10T10:00:00Z'));
+
+			await rm(tempDir, { recursive: true });
+		});
```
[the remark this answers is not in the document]
ben-vargas: Addressed in commit 9e85d53.

ryoppippi:
Thank you for implementing 
can you review this @Richard-Weiss

src/data-loader-deduplication.test.ts
```diff
@@ -0,0 +1,337 @@
...
+			};
+
+			const hash = createUniqueHash(data);
+			expect(hash).toBeNull();
+		});
+	});
+
+	describe('getEarliestTimestamp', () => {
+		let tempDir: string;
+
```
ryoppippi: please use fs-fixture for mocking files

ryoppippi (changes requested):

src/data-loader-deduplication.test.ts
ben-vargas: No problem, refactored in b5e8b35

ryoppippi: oh refer how to use fs-fixture in other test file. you can use `await using` syntax 
instead of using `afterEach` `beforeEach`

https://github.com/privatenumber/fs-fixture/tree/master?tab=readme-ov-file#using-keyword-explicit-resource-management

ryoppippi: https://github.com/ryoppippi/ccusage/blob/85afe7ba801978474356f9a5a16e62543188c584/src/debug.test.ts#L8-L30

ryoppippi (approved):
LGTM

ryoppippi:
Thanks

skylinesales (approved):

Issue #66, opened 2025-06-12
Support for `$CLAUDE_CONFIG_DIR`


ciffelia:
Claude Code uses the `$CLAUDE_CONFIG_DIR` environment variable to specify the path to the `.claude` directory, though this seems to be an undocumented feature. It would be convenient if `ccusage` could automatically recognize and utilize this variable.

Currently, I have to manually specify the path like this:

```
pnpx ccusage@0.8.0 --path "$CLAUDE_CONFIG_DIR"
```

ryoppippi:
It looks better approach. 
Send us a PR

PR #70, opened 2025-06-14
feat!: specify claude dir with `CLAUDE_CONFIG_DIR` and remove `-p`/`--path` option

timrogers:
## Summary

- **Remove -p short option**: Removed the \`-p\` short option since Claude data directory can now be configured via \`CLAUDE_CONFIG_DIR\` environment variable
- **Remove --path option entirely**: Completely removed the \`--path\` CLI option to simplify the interface and encourage environment variable usage
- **Update all commands**: Modified daily, monthly, session, and mcp commands to use \`getDefaultClaudePath()\` directly
- **Update documentation**: Updated README to remove all references to \`--path\` option and clarify environment variable usage

### CLI Interface Changes
- Removed \`path\` argument from \`shared-args.internal.ts\`
- Updated all command files to import and use \`getDefaultClaudePath()\` directly
- No longer accepts \`--path\` or \`-p\` options - users must use \`CLAUDE_CONFIG_DIR\` environment variable

### Documentation Updates
- Removed \`--path\` option from CLI options documentation in README
- Updated usage examples to remove \`--path\` references
- Clarified that only \`CLAUDE_CONFIG_DIR\` environment variable is supported for custom paths

### Test Coverage
- All existing tests continue to pass
- No test changes needed as the functionality remains the same, just configured differently

## Breaking Change Notice

⚠️ **This is a breaking change**: The \`--path\` and \`-p\` options are no longer available. Users who were using these options should set the \`CLAUDE_CONFIG_DIR\` environment variable instead:

\`\`\`bash
# Before (no longer works):
ccusage daily --path /custom/path

# After:
export CLAUDE_CONFIG_DIR="/custom/path"
ccusage daily
\`\`\`

## Test Plan

- [x] All existing tests pass
- [x] Linting passes
- [x] CLI works with default path (~/.claude)
- [x] CLI works with CLAUDE_CONFIG_DIR environment variable
- [x] All commands (daily, monthly, session, mcp) use environment variable correctly
- [x] Documentation updated to reflect changes

## Summary by CodeRabbit

- **New Features**
	- The tool now supports specifying the Claude data directory via the CLAUDE_CONFIG_DIR environment variable.
- **Bug Fixes**
	- Improved validation to ensure the specified data directory exists and is valid.
- **Documentation**
	- Updated instructions to reflect the use of the CLAUDE_CONFIG_DIR environment variable and removed references to the deprecated --path option.
- **Chores**
	- Added a new dependency to support directory validation.
- **Tests**
	- Added comprehensive tests for environment variable handling and directory validation.


ryoppippi:
Thank you!
I request two things

- it is ok to remove `-p` option!
- please use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) for your commit request. 
- Please use `feat!:` to remove `-p` option because we use [changelogithub](https://github.com/antfu/changelogithub) to manage release note

README.md
```diff
@@ -216,7 +228,7 @@ All commands support the following options:
 
 - `-s, --since <date>`: Filter from date (YYYYMMDD format)
 - `-u, --until <date>`: Filter until date (YYYYMMDD format)
-- `-p, --path <path>`: Custom path to Claude data directory (default: `~/.claude`)
+- `-p, --path <path>`: Custom path to Claude data directory (default: `$CLAUDE_CONFIG_DIR` or `~/.claude`)
```
ryoppippi: you can remove `-p` option

src/data-loader.test.ts
```diff
@@ -29,6 +30,69 @@ describe('formatDate', () => {
 	});
 });
 
+describe('getDefaultClaudePath', () => {
+	test('returns CLAUDE_CONFIG_DIR when environment variable is set', () => {
+		const originalEnv = process.env.CLAUDE_CONFIG_DIR;
+		const testPath = '/custom/claude/path';
+
+		try {
+			process.env.CLAUDE_CONFIG_DIR = testPath;
```
ryoppippi: because bun test is not isolated each other you need to mock `process.env` in each test case

src/data-loader.ts
```diff
@@ -13,7 +14,8 @@ import {
 import { groupBy } from './utils.internal.ts';
 
 export function getDefaultClaudePath(): string {
-	return path.join(homedir(), '.claude');
+	const envPath = process.env.CLAUDE_CONFIG_DIR;
+	return (envPath != null && envPath.trim() !== '') ? envPath : path.join(homedir(), '.claude');
```
ryoppippi:
```suggestion
	const defaultClaudeCodePath = path.join(homedir(), '.claude');
	const envClaudeCodePath = process.env.CLAUDE_CONFIG_DIR?.trim();
	if(existsSync(envClaudeCodePath)){
		return envClaudeCodePath
	}
	return defaultClaudeCodePath;
```

I think this is more readable! (you need to import `fs.existsSync` tho), and format it.
or, you may use `isDirectorySync` from https://github.com/sindresorhus/path-type to check if the path is directory	(I prefer this approach in this case over fs.existsSync)

src/data-loader.test.ts
ryoppippi: Also if you can mock claude project path using `fs-fixture` that'd be brilliant

ryoppippi:
@timrogers if you don't have time i'll do that

ryoppippi:
fixes: https://github.com/ryoppippi/ccusage/issues/66

PR #129, opened 2025-06-20
fix: use getClaudePaths for multi-directory support

nicobailon:
- Remove claudePath parameter from data loading functions in all commands
- This allows the data loader to check multiple Claude directories automatically
- Supports both ~/.config/claude and ~/.claude paths
- Fixes 'No Claude usage data found' error when data exists in alternate locations


ryoppippi (approved):
LGTM 
good PR 
thanks

PR #164, opened 2025-06-22
Fix for live usage tokens not getting updated 

dalmia007:
## Summary by CodeRabbit

- **Style**
  - Improved formatting and readability of live monitoring display functions. No changes to app behavior or output.

PR #168, opened 2025-06-22
Fix for usage not updating in live block view

dalmia007:
Related to https://github.com/ryoppippi/ccusage/pull/164

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved live monitoring to ensure data is always refreshed for each update cycle.


ryoppippi:
Ok so let's try it and see the result.
and let's debug the logic later

dalmia007:
Yes

Issue #59, opened 2025-06-11
Hitting error on installation


0xRaduan:
Error:
```
bunx ccusage daily
file:///private/tmp/bunx-501-ccusage@latest/node_modules/ccusage/dist/data-loader-nOFcMg_V.js:3156
				if ("function" != typeof o) throw new TypeError("Object is not disposable.");
				                                  ^

TypeError: Object is not disposable.
    at using (file:///private/tmp/bunx-501-ccusage@latest/node_modules/ccusage/dist/data-loader-nOFcMg_V.js:3156:39)
    at loadDailyUsageData (file:///private/tmp/bunx-501-ccusage@latest/node_modules/ccusage/dist/data-loader-nOFcMg_V.js:3295:31)
    at async Object.run (file:///private/tmp/bunx-501-ccusage@latest/node_modules/ccusage/dist/index.js:2895:21)
    at async executeCommand (file:///private/tmp/bunx-501-ccusage@latest/node_modules/ccusage/dist/index.js:1141:2)
    at async cli (file:///private/tmp/bunx-501-ccusage@latest/node_modules/ccusage/dist/index.js:1053:2)
    at async file:///private/tmp/bunx-501-ccusage@latest/node_modules/ccusage/dist/index.js:3267:1

Node.js v18.17.0
```

Same on npx:
```
npx ccusage@latest
Need to install the following packages:
  ccusage@0.7.0
Ok to proceed? (y) y
file:///Users/raduan/.npm/_npx/b8bc0fb451ae8722/node_modules/ccusage/dist/data-loader-nOFcMg_V.js:3156
				if ("function" != typeof o) throw new TypeError("Object is not disposable.");
				                                  ^

TypeError: Object is not disposable.
    at using (file:///Users/raduan/.npm/_npx/b8bc0fb451ae8722/node_modules/ccusage/dist/data-loader-nOFcMg_V.js:3156:39)
    at loadDailyUsageData (file:///Users/raduan/.npm/_npx/b8bc0fb451ae8722/node_modules/ccusage/dist/data-loader-nOFcMg_V.js:3295:31)
    at async Object.run (file:///Users/raduan/.npm/_npx/b8bc0fb451ae8722/node_modules/ccusage/dist/index.js:2895:21)
    at async executeCommand (file:///Users/raduan/.npm/_npx/b8bc0fb451ae8722/node_modules/ccusage/dist/index.js:1141:2)
[...]

PR #180, opened 2025-06-23
chore: set Node.js engine requirement to >=20 in package.json

aalimovs:
Symbol.dispose and Symbol.asyncDispose were shipped in 18.18 and required to run ccusage. See #59

## Summary by CodeRabbit

- **Chores**
  - Updated requirements to ensure the app runs on Node.js version 20.0.0 or higher.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,6 @@
 		"simple-git-hooks"
 	],
 	"engines": {
-		"node": ">=18.18.0"
-	}
+        "node": ">=20.0.0"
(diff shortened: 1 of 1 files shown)


ryoppippi:
@aalimovs oh yeah it is important to add `engine` property!!
But as we wrote in the document, we support `>node 20` because 18 is already deprecated

ryoppippi (approved):
LGTM

PR #201, opened 2025-06-26
feat: add --color and --no-color CLI flags for output control

gfx:
They are handled by `picolors`, so configuring `color` option is for users.


ryoppippi (changes requested):
@gfx
Thank you for your contribution.
It looks really nice, and I didn't know this spec of picocolors

- could you add a comment including ref url on this options so that contributors notice it
- could you update documentation site

src/_shared-args.ts
```diff
@@ -78,6 +78,12 @@ export const sharedArgs = {
 		description: 'Use cached pricing data for Claude models instead of fetching from API',
 		default: false,
 	},
+	color: { // handled by picocolors
+		type: 'boolean',
+		negatable: true,
+		description: 'Enable/disable colored output (default: auto)',
```
ryoppippi: you can define `noColor` option instead because `picocolors` doesn't support `--color` option
also remove `(default: auto)` and let gunshi show default automatically

src/_shared-args.ts
```diff
@@ -78,6 +78,12 @@ export const sharedArgs = {
 		description: 'Use cached pricing data for Claude models instead of fetching from API',
 		default: false,
 	},
```
ryoppippi, line 80: add comment about what is NO_COLOR

src/_shared-args.ts
gfx: `picocolors` does support `--color`: https://github.com/alexeyraspopov/picocolors/blob/main/picocolors.js

Clarified: 26fe689f771baa6baeef1c174a40ea183be735f8

gfx:
* Mentioned to FORCE_COLOR=1 and NO_COLOR=1
* Update the doc to mention --color and --no-color

gfx:
There's a bug that `ccusage | cat` (or `ccusage | pbcopy` in a macOS terminal) has partial colors, while `ccusage --no-color` does not. However, it's a separate problem from this PR. I'm willing to address it in another PR.

src/_shared-args.ts
ryoppippi: @gfx ok I didn't notice that `--color` option!!

ryoppippi:
@gfx Thank you for your all work!
My question is can we make a release after this PR? or should we wait for another PR from you?

gfx:
No need to wait for my another PR. It's a trivial bug fix!

ryoppippi (approved):
LGTM
Thanks for your works!!

PR #231, opened 2025-07-02
feat: add CSV output support with refactored shared handlers

watany-dev:
## Function 
 The `--csv` option allows csv output. This is an option that works well with other system integration and spreadsheets such as excel.

## Caution.

You may choose not to implement this in ccusage because it is the same as this.
```bash 
npx ccusage --json | jq -r ' 
 ["date", "inputTokens", "outputTokens", "cacheCreationTokens", "cacheReadTokens", "totalTokens", "totalCost ", "modelsUsed"], 
 (.daily[] | [ 
.date, 
.inputTokens, 
.outputTokens, 
.cacheCreationTokens, 
.cacheReadTokens, 
.totalTokens, 
.totalCost, 
 (. modelsUsed | join(";")) 
 ]) 
 | @csv 
' 
```


watany-dev:
Thanks for the prompt review. It is important to keep the code base clean and we are happy with the decision!

PR #238, opened 2025-07-03
refactor: Improve the error handling of the PricingFetcher class

Karibash:
## Overview

Improved error handling in the `PricingFetcher` class to implement more robust and consistent error processing. By leveraging the `Result` type from `@praha/byethrow`, error propagation and handling have become safer and more predictable.

## Key Changes

- Introduction of the Result pattern: All asynchronous operations are wrapped in the Result type for explicit error handling
- Enhanced type safety: Changed method return types to `R.ResultAsync<T, E>`
- Error propagation: Adopted a functional programming style for error handling using `R.pipe()`
- Other updates: Refactored processes dependent on the `PricingFetcher` class


src/pricing-fetcher.ts
```diff
@@ -49,97 +49,94 @@ export class PricingFetcher implements Disposable {
...
+						const pricing = new Map<string, ModelPricing>();
+						for (const [modelName, modelData] of Object.entries(data)) {
+							if (typeof modelData === 'object' && modelData !== null) {
+								const parsed = modelPricingSchema.safeParse(modelData);
+								if (parsed.success) {
+									pricing.set(modelName, parsed.data);
+								}
+								// Skip models that don't match our schema
+							}
+						}
```
ryoppippi, line 122: Amazing how to use pipe! Thanks!

ryoppippi:
@Karibash I just fix to use `Result` instead of `R`. I think when we use `R` we can reduce bundle size slightly. However using `Result` is more readable and LLM friendly! If I change my mind I probably start using `R`. anyway 勉強になりました！

PR #329, opened 2025-07-20
perf: optimize file reading with mtime filtering (up to 10x faster)

mbailey:
When using --since flag, pre-filter files based on modification time before reading contents. This provides significant performance improvement for users with extensive usage history.

Changes:
- Add mtime pre-filtering to skip files modified before --since date
- Optimize getEarliestTimestamp to stream only first 10 lines
- Add comprehensive documentation explaining why this is safe

Performance: up to 10x faster when loading data with --since flag

Why this is safe: Claude Code is append-only, so mtime always reflects when data was last added. If mtime < since date, the file genuinely has no data after that date.


mbailey:
> @mbailey thanks!! it looks lit! I need to check it out carefully

Thanks! Let me know if you want me to review it against the auto responses above. Happy to do what's required to reduce maintainer effort.

I love the program and appreciate you sharing it with other Claude users.

ryoppippi:
Hi @mbailey I'm reading this PR right now.
i have a question, why you don't use iterators

Issue #375, opened 2025-07-28
Adjust rate-limit projections in accordance with new weekly policy


JustinStitt:
Anthropic is rolling out a new rate-limiting policy with adjusted time frames.

Usage projection logic will need to change to more accurately inform users of their Claude Code usage.

See official Anthropic reddit post: https://www.reddit.com/r/ClaudeAI/comments/1mbo1sb/updating_rate_limits_for_claude_subscription/

ryoppippi:
@JustinStitt thanks.
I checked this and TBH I got shocked.
I don't have confidence to follow this complicated limitation. I need help from contributors!!!

PR #378, opened 2025-07-29
Add weekly usage command

juicemia:
This adds a weekly usage command. (https://github.com/ryoppippi/ccusage/issues/375) From what I read, it wasn't exactly clear how they were going to set the usage window so I added code that makes it work for both Sunday and Monday. When we know how the windows are going to work we can update it. I tested both though and they both seem to work.

With it set to Monday:
Screenshot 2025-07-29 at 11 06 14 AM

With it set to Sunday:
Screenshot 2025-07-29 at 11 06 45 AM

For now I left it set to Monday because I think that'll most likely be the day they reset it since it's the first day of the working week.

## Summary by CodeRabbit

* **New Features**
  * Added weekly usage reporting, allowing users to view aggregated token usage and costs by week.
  * Introduced a new CLI command to generate weekly usage reports with options for JSON or table output, filtering, and detailed breakdowns.
  * Enhanced data aggregation to support grouping usage by arbitrary time buckets (weekly or monthly).

* **Bug Fixes**
  * Ensured tests run consistently in the UTC timezone.

* **Chores**
  * Updated development dependencies to earlier versions for improved compatibility.


dv8silencer:
_(Question regarding how Anthropic actually does it)_
Question -- is it like the 5 hour thing where the 5 hour window starts the first time you use it (**or at least that is what I've seen people say about the 5hr window**)? So the "week" usage "metering" would start whenever the user first starts using Claude? Or does it start at a particular calendar time? If I started for the first time on Wednesday for example, then I wonder if it 7*24 hours from then.

package.json
```diff
@@ -51,7 +51,7 @@
 		"prepare": "simple-git-hooks",
 		"release": "bun lint && bun typecheck && vitest run && bun run build && bumpp",
 		"start": "bun run ./src/index.ts",
-		"test": "vitest",
+		"test": "TZ=UTC vitest",
```
ryoppippi, line 54: nice work

juicemia: 🙏

src/data-loader.ts
```diff
@@ -1045,31 +1091,76 @@ export async function loadSessionData(
...
+		})));
+}
+
+/**
+ * @param date - The date to get the week for
+ * @returns The date of the first day of the week for the given date
+ */
+function getDateWeek(date: Date): WeeklyDate {
+	// Anthropic is rolling this out on August 28, 2025. We're assuming that's when the
+	// weekly window is starting. Using Luxon here allows the code to easily account for
```
ryoppippi: luxon is too large, 
we need alternative

.gitignore
```diff
@@ -36,3 +36,5 @@ report.[0-9]_.[0-9]_.[0-9]_.[0-9]_.json
 .claude
 
 .eslintcache
+
+.vscode
```
ryoppippi: stop adding .vscode here wtf

ryoppippi (approved):
LGTM

.gitignore
JustinStitt: note to future travelers: add `.vscode/` to your `.git/info/exclude` file -- that way we don't pollute gitignores with ide-specific stuff.

Issue #433, opened 2025-08-06
Daily date column values off by one when wrapped to two lines in compact format


travisjhicks:
The daily date column can be off by one depending on terminal width. For example, today is the the 6th, but `ccusage` shows the 5th when the date is shown in its compact two-line format:

```
├───────────┼───────────────────┼──────────┼──────────┼────────────┤
│ 2025      │ - opus-4          │      216 │   10,416 │      $8.46 │
│ 08-05     │                   │          │          │            │
├───────────┼───────────────────┼──────────┼──────────┼────────────┤
│ Total     │                   │   34,741 │  580,973 │    $314.38 │
└───────────┴───────────────────┴──────────┴──────────┴────────────┘
```

When the window width is increased slightly, the date is output on one line with the correct date:

```
├────────────┼────────────────────┼───────────┼───────────┼─────────────┤
│ 2025-08-06 │ - opus-4           │       216 │    10,416 │       $8.46 │
├────────────┼────────────────────┼───────────┼───────────┼─────────────┤
│ Total      │                    │    34,741 │   580,973 │     $314.38 │
└────────────┴────────────────────┴───────────┴───────────┴─────────────┘
```

loadDailyUsageData does UTC to local YYYY-MM-DD:
`2025-08-06T16:00:00.000Z` → `2025-08-06` (for me)

formatDateCompact interprets the YYYY-MM-DD as UTC but it is already local time:
`2025-08-06` is interpreted as UTC (`2025-08-06T00:00:00Z`), which is `2025-08-05 20:00:00` in my local time. That gets displayed as `2025-08-05` even though it started out in local time.

PR #434, opened 2025-08-06
Fix off-by-one compact date formatting

travisjhicks:
Suggested fix for issue #433.

formatDateCompact is being given a date in YYYY-MM-DD format that has already been converted to the local time zone. Constructing a new Date from that string interprets it as UTC, which causes an off-by-one error for users at a negative offset from UTC.


ryoppippi (approved):
LGTM

PR #477, opened 2025-08-11
feat: add --id option to session command for custom statusline integration

Melvynx:
## Summary

This PR adds a `--id` option to the `session` command that exposes the `loadSessionUsageById` functionality through the CLI. This allows users to create custom statusline implementations without duplicating the existing logic.

## Motivation

The current built-in statusline command has limitations and isn't flexible enough for custom use cases. By exposing the underlying session lookup functionality, users can:

- Create their own custom statuslines with specific formatting
- Build scripts that query specific session data programmatically  
- Integrate session usage data into other tools and workflows
- Avoid duplicating the complex session loading logic

Currently, the `loadSessionUsageById` function exists and is used internally by the statusline command, but there's no way for users to access this functionality directly. This change makes it available through the CLI interface.

## Changes

- ✅ Add `--id/-i` option to session command
- ✅ Support both JSON and table output formats for specific sessions
- ✅ Maintain full backward compatibility with existing session listing
- ✅ Proper error handling for non-existent session IDs
- ✅ Support for jq processing with `--jq` flag

## Usage Examples

```bash
# Get session data in table format
ccusage session --id <session-id>

# Get session data in JSON format  
ccusage session --id <session-id> --json

# Process with jq for custom formatting
ccusage session --id <session-id> --json --jq '.totalCost'

# Use in custom statusline scripts
COST=$(ccusage session --id "$SESSION_ID" --json --jq '.totalCost')
echo "Current session cost: \$${COST}"
```

## Why This Makes Sense

Instead of having users reverse-engineer or duplicate the session loading logic, this change makes the already-existing `loadSessionUsageById` function accessible via CLI. This promotes code reuse and enables better custom integrations while keeping the core logic centralized.

The built-in statusline is quite basic and not very customizable. This change enables users to build much more sophisticated and personalized statuslines tailored to their specific needs, without having to reimplement the complex session data loading and cost calculation logic.

## Technical Implementation

- The implementation reuses the existing `loadSessionUsageById` function from `data-loader.ts`
- Added proper TypeScript types and error handling
- Maintains the same cost calculation modes (`auto`, `calculate`, `display`)
- Supports offline mode for faster performance
- Full integration with existing CLI framework and shared arguments

## Summary by CodeRabbit

- New Features
  - Added per-session lookup: query a single session by ID using --id or -i. Works with JSON output and jq filtering. Returns a clear result when a session isn’t found. Existing session listing remains unchanged.

- Documentation
  - Added “Specific Session Lookup” guide with examples (table view, JSON, jq field extraction, statusline usage).
  - Clarified session ID format and how to locate IDs.
  - Expanded “Command Options” with ID lookup details and use cases.
  - Added date filtering examples (e.g., --since, --until, relative dates).


ryoppippi:
@Melvynx 
thanks! but could you remove the pnpm lock file?

Melvynx:
hey @ryoppippi it's done and i clean the eslint errors ;)

src/commands/session.ts
```diff
@@ -10,21 +12,175 @@ import {
...
+			+ usage.output_tokens
+			+ (usage.cache_creation_input_tokens ?? 0)
+			+ (usage.cache_read_input_tokens ?? 0);
+	}, 0);
+}
+
+/**
+ * Handles session ID lookup and output formatting
+ */
+async function handleSessionIdLookup(
```
ryoppippi: do not add logic in command script

ryoppippi (changes requested):

ryoppippi (approved):
LGTM

PR #497, opened 2025-08-13
Fix: Windows hang when CWD contains symbolic links

FallDownTheSystem:
## Problem

ccusage would hang indefinitely on Windows when run from directories containing symbolic links (junction points) such as:
- `C:\Users\{username}\Documents` (contains My Music, My Pictures, My Videos junctions)
- `C:\Users\{username}` (contains My Documents, Application Data, etc.)

This issue was discovered through https://github.com/Owloops/claude-powerline which uses ccusage's `loadSessionUsageById()` function and would hang when run from these directories.

## Root Cause

The issue is ultimately caused by a bug in tinyglobby where Windows absolute paths with backslashes (e.g., `C:\Users\...\**\file.jsonl`) would cause the glob function to incorrectly process the path when the current working directory contained symbolic links.

## Solution

Convert backslashes to forward slashes in glob patterns. Forward slashes are cross-platform compatible and avoid the tinyglobby bug entirely.

## Changes

- Modified `loadSessionUsageById` in `src/data-loader.ts` to replace backslashes with forward slashes in the glob pattern

## Testing

Tested the following scenarios on Windows:
- ✅ Running from directories WITH symlinks (Documents, Home) - now works correctly
- ✅ Running from directories WITHOUT symlinks - continues to work
- ✅ Unix/Linux/macOS compatibility - unaffected (already use forward slashes)

## Technical Details

The bug manifests when:
1. Using Windows absolute paths with backslashes (`C:\...`)
2. Current working directory contains symbolic links/junction points
3. tinyglobby attempts to process the path

By converting to forward slashes, we maintain full cross-platform compatibility while avoiding the tinyglobby bug.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Windows path handling for session-usage discovery by normalizing glob patterns to use forward slashes; session usage data is now reliably found (including projects accessed via symlinked directories). No changes to commands, configuration, or outputs; behavior on macOS/Linux unchanged.

### DIFF
--- a/src/data-loader.ts
+++ b/src/data-loader.ts
@@ -1280,9 +1280,9 @@ export async function loadSessionUsageById(
 	const claudePaths = getClaudePaths();
 
 	// Find the JSONL file for this session ID
-	// On Windows, replace backslashes from path.join with forward slashes for tinyglobby compatibility
+	// Use path.posix to ensure forward slashes in pattern for tinyglobby compatibility
 	const patterns = claudePaths.map(p => 
-		path.join(p, 'projects', '**', `${sessionId}.jsonl`).replace(/\\/g, '/')
+		path.posix.join(p, 'projects', '**', `${sessionId}.jsonl`)
 	);
 	const jsonlFiles = await glob(patterns);
 
--- a/src/data-loader.ts
+++ b/src/data-loader.ts
@@ -1280,7 +1280,10 @@ export async function loadSessionUsageById(
 	const claudePaths = getClaudePaths();
 
 	// Find the JSONL file for this session ID
-	const patterns = claudePaths.map(p => path.join(p, 'projects', '**', `${sessionId}.jsonl`));
+	// Convert backslashes to forward slashes to avoid tinyglobby bug with Windows paths in directories containing symlinks
+	const patterns = claudePaths.map(p => 
+		path.join(p, 'projects', '**', `${sessionId}.jsonl`).replace(/\\/g, '/')
+	);
 	const jsonlFiles = await glob(patterns);
 
 	if (jsonlFiles.length === 0) {
--- a/src/data-loader.ts
+++ b/src/data-loader.ts
@@ -1280,7 +1280,7 @@ export async function loadSessionUsageById(
 	const claudePaths = getClaudePaths();
 
 	// Find the JSONL file for this session ID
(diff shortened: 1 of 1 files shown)


src/data-loader.ts
```diff
@@ -1280,7 +1280,10 @@ export async function loadSessionUsageById(
 	const claudePaths = getClaudePaths();
 
 	// Find the JSONL file for this session ID
-	const patterns = claudePaths.map(p => path.join(p, 'projects', '**', `${sessionId}.jsonl`));
+	// Convert backslashes to forward slashes to avoid tinyglobby bug with Windows paths in directories containing symlinks
+	const patterns = claudePaths.map(p => 
+		path.join(p, 'projects', '**', `${sessionId}.jsonl`).replace(/\\/g, '/')
```
ryoppippi: @FallDownTheSystem why you use this greps?
```
/\\/g
```

FallDownTheSystem: It's a common way to replace all instances, you could also do `replaceAll('\\', '/')`

But as it turns out, tinyglobby doesn't support backwards slashes in patterns, as the author kindly let me know in the above issue.

FallDownTheSystem: I improved the comment now that I understand the issue better. And just to clarify, `path.join` produces paths with backwards slashes on Windows. Using `path.posix.join`, which outputs forward slashes regardless of OS would also be a suitable alternative.

ryoppippi: thanks! I prefer usnig `path.posix.join` because it looks more mature way I think. also what happens if some pathes in linux includes `\` character. just curios.

FallDownTheSystem: Updated, With path.posix.join it should keep backwards slashes in the path if linux paths contain them.

ryoppippi:
@FallDownTheSystem hi I think this PR worth opening again! I'll approve this pr.
what do you think?

ryoppippi:
I'll open it again. if you don't mind i'll merge it! tell me your thoughts!

FallDownTheSystem:
That's fine, the project that used ccusage no longer uses it. Coderabbitai brought up a good point, you don't want to use path.posix.join as it won't remove backslashes from the Windows paths that already exist.

ryoppippi (approved):
LGTM

ryoppippi (approved):
LGTM

PR #613, opened 2025-09-02
fix!: remove session order flag

### LINKED ISSUE #350: --desc flag not working
Image

--desc flag (as I understand it) should show most expensive first in the table? This doesn't appear to be happening

amorriscode:
Resolves #350

## Summary by CodeRabbit

- Refactor
  - Simplified the session command by removing the explicit --order flag; sessions now use the default sort automatically.

- Documentation
  - Updated Getting Started: renamed “Find Expensive Sessions” to “Analyze Sessions” and simplified the session command examples.
  - Updated Session Reports: removed the “Sort Order” section and revised examples to reflect default sorting (no --order flag).
  - Ensured examples consistently use ccusage session without explicit ordering flags.

### DIFF
--- a/config-schema.json
+++ b/config-schema.json
@@ -483,16 +483,6 @@
 									"markdownDescription": "Number of sample discrepancies to show in debug output (default: 5)",
 									"default": 5
 								},
-								"order": {
-									"type": "string",
-									"enum": [
-										"desc",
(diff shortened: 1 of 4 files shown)


ryoppippi:
@amorriscode Thanks!

Issue #626, opened 2025-09-12
Is it possible to support codex?


random-yang:

FutureExcited:
would be amazing

ryoppippi:
Guys don't forget to support us!

PR #641, opened 2025-09-17
feat: codex

ryoppippi:
codex-cli

Fixes: https://github.com/ryoppippi/ccusage/issues/626

## Summary
- introduce a monthly Codex usage report that mirrors the existing daily command (shared args, pricing source, responsive table output)
- extract shared command helpers so daily/monthly stay in sync and reuse the shared pricing/formatting logic
- document the new command in the package README and Codex guide (daily page cross-link, new monthly page)
- add tsdown builds for @ccusage/internal and @ccusage/terminal so their d.ts files exist during app bundling

## Testing
- pnpm run format
- pnpm typecheck
- pnpm run test
- pnpm --filter @ccusage/internal build
- pnpm --filter @ccusage/terminal build
- pnpm --filter ccusage build

## Summary by CodeRabbit

- **New Features**
  - Introduced Codex CLI (Beta): daily & monthly commands with JSON/table output, per‑model breakdowns, totals, timezone/locale options, offline pricing, and CODEX_HOME config.

- **Refactor**
  - Centralized pricing into a shared, offline-capable pricing system; simplified local pricing fetchers and reduced direct pricing surface.
  - Packaging/manifest reorganized for ESM, consolidated engine constraints, and adjusted published exports.

- **Documentation**
  - Added Codex docs and an "Apps Are Bundled" guidance section.

- **Chores**
  - Added ESLint/build/test configs, new internal package, and assorted packaging/tooling updates.

### DIFF
--- a/apps/codex/CLAUDE.md
+++ b/apps/codex/CLAUDE.md
@@ -40,7 +40,7 @@
 | `reasoning_output_tokens`                       | Optional breakdown for reasoning            | Informational only; already included in `output_tokens`                  |
 | `total_tokens`                                  | Cumulative total emitted by Codex           | Used verbatim when present; legacy entries fall back to `input + output` |
 
-Parsing normalises every event through these rules. When we have to synthesise totals for legacy JSONL files we explicitly skip adding reasoning so the display matches what Codex billed. Events that rely on model/pricing fallbacks carry `isFallbackModel === true`, and aggregated model rows expose `isFallback` so table/JSON output highlights the assumption.
+Parsing normalizes every event through these rules. When we have to synthesize totals for legacy JSONL files we explicitly skip adding reasoning so the display matches what Codex billed. Events that rely on model/pricing fallbacks carry `isFallbackModel === true`, and aggregated model rows expose `isFallback` so table/JSON output highlights the assumption.
 
 ## CLI Usage
 


ben-vargas:
Nice seems to be working great. 

Haha... Codex is soooo much cheaper than Claude (especially Opus)!

CleanShot 2025-09-17 at 18 25 07@2x

ryoppippi:
Those warning are ok. We just build them. Also we use macro in ccusage as well. Don't worry!
So I just wanna make sure 
- how cache token works
- is it okay just use gpt5 pricing

ryoppippi:
oh bin path was wrong! i fixed it

ben-vargas:
I've been doing a bit of digging into the logic, I think there's a bug in the calculations... I believe that output_tokens already includes reasoning_output_tokens right? So if it calculates `output = output + reasoning` it would actually be double counting the reasoning tokens.

Code change to consider...

- In apps/codex/src/token-utils.ts:32-39 stop adding reasoning onto the output bucket. Replace `const outputTokens = usage.outputTokens + usage.reasoningOutputTokens;` with just `usage.outputTokens` so billing aligns with the raw log semantics (reasoning_output_tokens is already part of output_tokens). The rest of the formula (non‑cached vs cached) can stay as-is.

- Update the cost expectations in the in-source tests that currently assume reasoning is double billed:
    • apps/codex/src/token-utils.ts (calculateCostUSD test)
    • apps/codex/src/daily-report.ts and apps/codex/src/monthly-report.ts (they hard-code expected costs based on the old math). Recompute those numbers using the corrected formula.

- After the change, re-run bun ./src/index.ts daily --json (or the test suite) to make sure daily/monthly reports still match the raw log totals—only the USD column should drop slightly. If you need a sanity check, compare totalWith vs totalWithout from the earlier spot script; the latter is the value the CLI should now report.

- No other adjustments are necessary unless you want to tweak the table labels (current “Output” column still shows the full output_tokens, which already includes reasoning).

A bit of analysis from gpt-5-codex:

- In the raw Codex logs, total_tokens = input_tokens + output_tokens even when reasoning_output_tokens is non‑zero. Example (2025‑09‑14T06:48:05Z):
    input=12 460, output=1 042, reasoning=1 024, total=13 502.
    The delta output - reasoning = 18 lines up with the short assistant reply that was actually shown in the terminal. The extra 1 024 tokens are the hidden chain‑of‑thought that the
  model emitted, and they’re already counted inside output_tokens.
  - That pattern holds across the entire dataset. I scanned /home/code/.codex/sessions and found 2 015 reasoning events; in every single case output_tokens ≥ reasoning_output_tokens, and
  total_tokens stayed at input + output—never input + output + reasoning. So the recorder is already folding the reasoning stream into the output_tokens figure.
  - Because of that, the current cost helper (calculateCostUSD) is double counting: it feeds outputTokens + reasoningOutputTokens into the pricing calculation even though those reasoning
  tokens were already included in outputTokens. The right thing to do is bill on outputTokens alone (if you still want a separate “reasoning” column in reports, display it for insight but
  don’t add it again for billing).
  - In short: the comment from Grok is accurate for these logs—reasoning is a subset of output. The line in token-utils.ts that effectively does output += reasoning should be removed (and
  the reporting logic should subtract reasoning if you want the “Output” column to show only user‑visible tokens).

PR #665, opened 2025-09-24
feat(ccusage): add Claude service status command

him0:
## Summary
- Add new `status` subcommand to check Claude service status from status.claude.com API
- Display current Claude service status with color-coded indicators
- Support JSON output with `--json` flag

## Features
- **Color-coded status indicators**: 
  - Green: operational
  - Yellow: minor/degraded service  
  - Red: major/critical/outage
- **JSON output support**: `ccusage status --json`
- **Proper error handling**: Graceful failure with informative messages

## Usage
```bash
ccusage status                    # Show status with colors
ccusage status --json             # JSON output
```

```
$ pnpm run start status

> ccusage@17.0.2 start /Users/him0/src/ccusage/apps/ccusage
> bun ./src/index.ts status

Claude Status: Partially Degraded Service - https://status.claude.com
```
CleanShot 2025-09-24 at 18 02 29@2x

## Test plan
- [x] Test status display with current API response (Partially Degraded Service - Yellow)
- [x] Test JSON output format
- [x] Test error handling for API failures
- [x] Verify color coding works correctly
- [x] Confirm URL is dynamically retrieved from API response

## Summary by CodeRabbit

* **New Features**
  * Added a CLI command to show Claude service status with human-readable and JSON output modes, colorized health indicators, and non-zero exit on failures.

* **Configuration**
  * Introduced a public status configuration object (json: false by default). Available globally and as a per-command override; unknown keys are disallowed.

* **Documentation**
  * Added usage docs and guide entries describing the status command, outputs, and examples.

### DIFF
--- a/apps/ccusage/README.md
+++ b/apps/ccusage/README.md
@@ -95,6 +95,7 @@ npx ccusage daily    # Daily token usage and costs
 npx ccusage monthly  # Monthly aggregated report
 npx ccusage session  # Usage by conversation session
 npx ccusage blocks   # 5-hour billing windows
+npx ccusage status   # Check Claude service status
 npx ccusage statusline  # Compact status line for hooks (Beta)
 
 # Live monitoring
@@ -125,6 +126,7 @@ npx ccusage monthly --compact  # Compact monthly report
 - ⏰ **5-Hour Blocks Report**: Track usage within Claude's billing windows with active block monitoring
 - 📈 **Live Monitoring**: Real-time dashboard showing active session progress, token burn rate, and cost projections with `blocks --live`
 - 🚀 **Statusline Integration**: Compact usage display for Claude Code status bar hooks (Beta)
+- 🌐 **Claude Status**: Check Claude service operational status with `ccusage status`
 - 🤖 **Model Tracking**: See which Claude models you're using (Opus, Sonnet, etc.)
 - 📊 **Model Breakdown**: View per-model cost breakdown with `--breakdown` flag
 - 📅 **Date Filtering**: Filter reports by date range using `--since` and `--until`
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -94,6 +94,7 @@ export default defineConfig({
 						{ text: 'Library Usage', link: '/guide/library-usage' },
 						{ text: 'MCP Server', link: '/guide/mcp-server' },
 						{ text: 'JSON Output', link: '/guide/json-output' },
+						{ text: 'Claude Status', link: '/guide/status' },
 						{ text: 'Statusline Integration', link: '/guide/statusline' },
 					],
 				},
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -58,6 +58,10 @@ Unlike other CLI tools, we pay extreme attention to bundle size. ccusage achieve
 - Token limit warnings and projections
 - Automatic refresh with configurable intervals
 
+### 🌐 Service Status
+
+- **Claude Status** - Check current operational status of Claude services
+
 ### 🔧 Flexible Configuration
 
 - **JSON Configuration Files** - Set defaults for all commands or customize per-command
--- a/docs/guide/status.md
+++ b/docs/guide/status.md
@@ -1,0 +1,90 @@
+# Claude Status
+
+Check the current operational status of Claude services directly from the command line.
+
+## Basic Usage
+
+```bash
+ccusage status
+```
+
+## Example Output
+
+```
+Claude Status: All Systems Operational - https://status.claude.com
+```
+
+The status message is color-coded based on the current service state:
+
+| Color  | Status      | Description                              |
+| ------ | ----------- | ---------------------------------------- |
+| Green  | Operational | All systems are working normally         |
+| Yellow | Degraded    | Some services may be experiencing issues |
+| Red    | Outage      | Partial or major service outage          |
+
+## JSON Output
+
+Export status data as JSON for programmatic use:
+
+```bash
+ccusage status --json
+```
+
+```json
+{
+	"status": {
+		"description": "All Systems Operational",
+		"indicator": "none"
+	},
+	"page": {
+		"id": "...",
+		"name": "Claude",
+		"url": "https://status.claude.com",
+		"time_zone": "Etc/UTC",
+		"updated_at": "2025-01-15T12:00:00.000Z"
+	}
+}
+```
+
+### Status Indicators
+
+The `indicator` field in JSON output can be:
+
+- `none` - All systems operational
+- `minor` - Minor issues or degraded performance
+- `major` - Major outage affecting services
+- `critical` - Critical outage
+
+## Use Cases
+
+### Quick Status Check
+
+Before starting a coding session, verify Claude services are available:
+
+```bash
+ccusage status
+```
+
+### Scripting and Automation
+
+Use JSON output in scripts to check Claude availability:
+
+```bash
+# Check if Claude is operational
+if ccusage status --json | grep -q '"indicator": "none"'; then
+  echo "Claude is ready!"
+fi
+```
+
+### Troubleshooting
+
+If you're experiencing issues with Claude Code, check the service status first:
+
+```bash
+ccusage status
+# If not operational, visit the status page for more details
+```
+
+## Related Commands
+
+- [Statusline Integration](/guide/statusline) - Compact usage display for Claude Code status bar


apps/ccusage/src/commands/status.ts
```diff
@@ -0,0 +1,80 @@
...
+			else if (indicator === 'major' || indicator === 'critical' || description.toLowerCase().includes('outage')) {
+				styledStatus = pc.red(description);
+			}
+
+			log(`Claude Status: ${styledStatus} - ${status.page.url}`);
+		}
+	},
+});
+
+if (import.meta.vitest != null) {
```
ryoppippi: looks nice.
can we remove this test? because this test doesn't mean anything.
also i don't want to write test in each command(prob i wrote it in claude.md). if we implement some logic and need write a test, do it in other files like _claude-status-api.ts

apps/ccusage/src/_claude-status-api.ts
```diff
@@ -0,0 +1,48 @@
...
+	}),
+});
+
+export type ClaudeStatus = v.InferInput<typeof claudeStatusSchema>;
+
+/**
+ * Fetch Claude status from status.claude.com API
+ * @returns Result containing Claude status data or error
+ */
+export async function fetchClaudeStatus(): Promise<{ success: true; value: ClaudeStatus } | { success: false; error: Error }> {
```
ryoppippi: and i need a test for this method

apps/ccusage/src/commands/status.ts
```diff
@@ -0,0 +1,80 @@
...
+			let styledStatus = description;
+			if (indicator === 'none' || description.toLowerCase().includes('operational')) {
+				styledStatus = pc.green(description);
+			}
+			else if (indicator === 'minor' || description.toLowerCase().includes('degraded')) {
+				styledStatus = pc.yellow(description);
+			}
+			else if (indicator === 'major' || indicator === 'critical' || description.toLowerCase().includes('outage')) {
+				styledStatus = pc.red(description);
+			}
```
ryoppippi: i think this is a "logic", so it's better to create a method and define union string in _claude-status-api file. 
ofc we need a test for this logic

apps/ccusage/src/_claude-status-api.ts
```diff
@@ -0,0 +1,48 @@
...
+
+		// Validate response data
+		const validatedData = v.parse(claudeStatusSchema, data);
+		return { success: true, value: validatedData };
+	}
+	catch (error) {
+		const err = error instanceof Error ? error : new Error(String(error));
+		return { success: false, error: err };
+	}
+}
```
ryoppippi: also please use byethrow instead of implementing your own result type

ryoppippi: and i think we don't need headers

ryoppippi: and if you want to return some result function, use `v.safeParse` instead of `v.parse`

ryoppippi: i think we have byethrow mcp and we've used byethrow in many places, so claude/codex can write it so well

apps/ccusage/src/_claude-status-api.ts
```diff
@@ -0,0 +1,48 @@
...
+		indicator: v.string(),
+	}),
+	page: v.object({
+		id: v.string(),
+		name: v.string(),
+		url: v.string(),
+		time_zone: v.string(),
+		updated_at: v.string(),
+	}),
+});
```
ryoppippi: this schema itself is so good!
BUT, we don't need to export this schema because we don't use it in other files.

same as infered type

him0:
@ryoppippi Thank you for the comment.
I’d like to hear your opinion on the naming convention of the subcommands and on whether the feature itself is necessary.
If the overall direction seems good, I’ll proceed with addressing the comment.

apps/ccusage/src/commands/status.ts
```diff
@@ -22,20 +22,21 @@ export const statusCommand = define({
...
-					error: statusResult.error.message,
+					error: error.message,
 					success: false,
 				}));
 			}
 			else {
 				logger.error(errorMessage);
 			}
 
-			process.exit(1);
```
ryoppippi: keep process.exit(1)

apps/ccusage/src/_claude-status-api.ts
```diff
@@ -48,3 +89,87 @@ export async function fetchClaudeStatus(): Result.ResultAsync<ClaudeStatus, Erro
 
 	return result();
 }
+
+if (import.meta.vitest != null) {
+	const { describe, it, expect } = import.meta.vitest;
```
ryoppippi: we are using vitest global, so no need import those test methods

apps/ccusage/src/_claude-status-api.ts
```diff
@@ -93,6 +93,57 @@ export async function fetchClaudeStatus(): Result.ResultAsync<ClaudeStatus, Erro
...
+			// Verify Result type structure
+			if (Result.isSuccess(result)) {
+				expect(result).toHaveProperty('value');
+				expect(typeof result.value).toBe('object');
+			}
+			else {
+				expect(result).toHaveProperty('error');
+				expect(result.error).toBeInstanceOf(Error);
+			}
+		});
```
ryoppippi: it's not a good idea to use conditions or branches in test

ryoppippi (changes requested):

apps/ccusage/src/_claude-status-api.ts
```diff
@@ -0,0 +1,226 @@
...
+	indicator: string,
+	description: string,
+	enableColors?: boolean,
+): Formatter {
+	// Determine if colors should be enabled
+	const shouldColor = enableColors ?? pc.isColorSupported;
+
+	if (!shouldColor) {
+		return (text: string | number | null | undefined) => String(text ?? '');
+	}
```
ryoppippi: picocolors support NO_COLOR environment, so we don't need to add this condition. refer another files 
https://github.com/ryoppippi/ccusage/pull/201

apps/ccusage/src/_claude-status-api.ts
```diff
@@ -0,0 +1,226 @@
...
+
+if (import.meta.vitest != null) {
+	const { describe, it, expect } = import.meta.vitest;
+
+	describe('fetchClaudeStatus', () => {
+		it('should fetch and parse actual API response', async () => {
+			const result = await fetchClaudeStatus();
+
+			// Only run test if API is available
+			if (Result.isSuccess(result)) {
```
him0, line 101: These conditionals exist because we're making actual HTTP requests without mocking `fetch`. The tests would fail unpredictably based on network/API availability otherwise.
Should I implement fetch mocking with `vi.mock()` to eliminate these conditionals and make the tests deterministic? Happy to refactor if that's preferred.

apps/ccusage/src/_claude-status-api.ts
```diff
@@ -0,0 +1,229 @@
...
+
+			// Always verify that we get a Result type back
+			expect(Result.isSuccess(result) || Result.isFailure(result)).toBe(true);
+		});
+
+		it('should have proper structure when successful', async () => {
+			const result = await fetchClaudeStatus();
+
+			// Skip this test if API is not available, but don't use conditional assertions
+			if (Result.isFailure(result)) {
```
him0, line 109: These conditionals exist because we're making actual HTTP requests without mocking `fetch`. The tests would fail unpredictably based on network/API availability otherwise.
Should I implement fetch mocking with `vi.mock()` to eliminate these conditionals and make the tests deterministic? Happy to refactor if that's preferred.

him0:
@ryoppippi Thanks for the quick code review. I’ve addressed the comments you left.

That said, I’d really like your view on the overall direction:
	•	Do you think adding this status subcommand itself makes sense?
	•	Is the naming convention for subcommands acceptable?

Also, I’m not confident about the testing strategy. I can’t decide whether we should mock fetch or not, so I’d appreciate it if you could set the direction there as well.

Once we agree on these points, I’ll proceed accordingly.

ryoppippi:
@him0 
thank you! we'd love to add this subcommand! but we just want to keep the quality of code

apps/ccusage/src/_claude-status-api.ts
ryoppippi: we don't need to mock! we just remove condition and if the test fails we can realise that api has some trouble

him0:
@ryoppippi The updates are done. Can you review it once more?

ryoppippi:
sorry i was/am so busy. i'll take a look later. hopefully on weekend!!

Issue #666, opened 2025-09-24
Statusline context percentage incorrect


chrisbeardy:
When using the statusline, the context percentage is showing very low, for example, the context could be shown as 130,000 tokens but the percentage could be 13%.

chrisbeardy:
I've noticed its just dividing it by 1000. Looks like there is a maths error somewhere.

ryoppippi:
so what is the difference between claude-plan and claude model from api

chrisbeardy:
as of today this appears to have corrected itself, but this is only because litellm for sonnet 4.5 has max tokens of 200000

<img width="529" height="749" alt="Image" src="https://github.com/user-attachments/assets/3605deed-932f-4baa-bf41-2770d1116b40" />

willma-andela:
This is still an issue

<img width="621" height="311" alt="Image" src="https://github.com/user-attachments/assets/ad8d0b5f-9988-44bf-b734-b1e9b3812ea4" />

<img width="116" height="35" alt="Image" src="https://github.com/user-attachments/assets/832bd152-fb4a-46d3-a145-0279021fdb27" />

Would love a fix for this as I use this information to help me decide when to `/clear`

ryoppippi:
The current statusline still relies on hookData.context_window.total_input_tokens, which can diverge from /context and the actual current usage. That problem is covered by #823, and #923 is the active PR direction for fixing it, so I am closing this as a duplicate.

willma-andela:
@ryoppippi you closed #923, so I think this should be reopened.

PR #671, opened 2025-09-26
Statusline context window overrides

chrisbeardy:
a potential fix for #666 to allow user overrides. Entirely generated by Claude Code (I'm not a javascript/typescript dev) so please review. Also bun keeps crashing on my PC so can't update schema and don't have time to debug it.

## Summary by CodeRabbit

- New Features
  - Added CLI options to statusline: --contextWindowMode (auto, claude-api, claude-plan) and --customContextWindow to override size.
  - Config now supports defaults for contextWindowMode and customContextWindow.
  - Improved context window detection with automatic model-based lookup and sensible fallbacks, yielding more accurate usage percentages.
- Bug Fixes
  - More reliable handling of context limits when offline or model data is unavailable, reducing incorrect usage calculations.

### DIFF
--- a/apps/ccusage/config-schema.json
+++ b/apps/ccusage/config-schema.json
@@ -104,6 +104,24 @@
 							"description": "Force compact mode for narrow displays (better for screenshots)",
 							"markdownDescription": "Force compact mode for narrow displays (better for screenshots)",
 							"default": false
+						},
+						"contextWindowMode": {
+							"type": "string",
+							"enum": [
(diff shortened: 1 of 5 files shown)


ryoppippi:
Closing this stale PR because the statusline context implementation has changed substantially since it was opened. Current main already prefers Claude Code hook context_window data and the remaining correctness issue is better represented by #923/#666, which use current_usage rather than adding manual context-window override modes.

PR #677, opened 2025-09-29
fix(ccusage): handle case-insensitive session ID lookup

### LINKED ISSUE #675: Bug: session cost calculation breaks with uppercase session-id
Image

Here, it throws an error message saying “No session found with ID” even though the session_id exists in the Claude project directory.

mgiovani:
## Summary

Fixes #675 - Session cost calculation was breaking when users provided uppercase session IDs because the file lookup used case-sensitive glob matching.

## Changes

- Updated `loadSessionUsageById` function to use case-insensitive glob matching by setting `caseSensitiveMatch: false`
- Added three comprehensive test cases:
  - Uppercase session ID with lowercase file
  - Lowercase session ID with uppercase file
  - Mixed-case session ID matching

## Solution Details

The fix leverages tinyglobby's built-in `caseSensitiveMatch` option (available since v0.2.14) to enable case-insensitive file matching. This allows session IDs to be looked up regardless of case, making the tool more user-friendly.

## Testing

All existing tests pass, plus three new test cases specifically for case-insensitive matching:

```bash
✓ loads session with uppercase ID when file is lowercase
✓ loads session with lowercase ID when file is uppercase  
✓ loads session with mixed-case ID
```

Ran full test suite:
- `pnpm run format` ✓
- `pnpm typecheck` ✓
- `pnpm test` ✓ (259 tests in ccusage, all passing)

## Impact

- **No breaking changes** - purely a bug fix
- **No performance impact** - same glob operation with different matching mode
- Affects: `session` command with `--id` flag and `statusline` command cost calculation

## Summary by CodeRabbit

* **Bug Fixes**
  * Session usage loading now works regardless of filename capitalization, addressing issues where sessions couldn't be found if the file name’s case didn’t match the requested ID. This improves cross-platform reliability (e.g., case-sensitive file systems) and reduces user friction when accessing session data. The change is fully backward-compatible and requires no user action.

### DIFF
--- a/apps/ccusage/src/data-loader.ts
+++ b/apps/ccusage/src/data-loader.ts
@@ -1093,7 +1093,7 @@ export async function loadSessionUsageById(
 	// Find the JSONL file for this session ID
 	// On Windows, replace backslashes from path.join with forward slashes for tinyglobby compatibility
 	const patterns = claudePaths.map(p => path.join(p, 'projects', '**', `${sessionId}.jsonl`).replace(/\\/g, '/'));
-	const jsonlFiles = await glob(patterns);
+	const jsonlFiles = await glob(patterns, { caseSensitiveMatch: false });
 
 	if (jsonlFiles.length === 0) {
 		return null;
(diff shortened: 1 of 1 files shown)


ryoppippi:
Thank you for the contribution and for documenting the session-id case-sensitivity problem clearly.

I am sorry that the Rust migration made this no longer mergeable as-is. The production CLI has moved from the TypeScript implementation to the Rust implementation under `rust/crates/ccusage`, and this PR targets the retired TypeScript runtime path. Because this PR was opened in 2025 and is no longer mergeable against the current `main`, I am closing it.

If this behavior is still needed on current `main`, please reopen it as a new Rust implementation or a focused issue against the Rust CLI. The underlying bug report is still useful context, but this TypeScript patch is obsolete now.

PR #768, opened 2025-12-25
feat(ccusage): add --no-cost flag to hide cost column

### LINKED ISSUE #625: Add an option to omit cost in output
I'm a Claude Pro Max subscriber. While it's kind of fun to see how much I would have spent on my usage if I was using the API, I'm more interested in seeing how my token use changes over time. The cost is distracting since it has a $ sign and is eye-grabbing.

This could either be a new flag like `--no-display-cost` to omit the cost column, or a new `--mode none` that skips cost calculation and output.

majiayu000:
## Summary

Add `--no-cost` option to all commands (daily, monthly, session, blocks) to hide cost information from table output. This is useful for Claude Pro subscribers who pay a flat rate and don't need cost tracking.

- Add `noCost` flag to shared args
- Update table utilities to support `hideCost` option
- Apply `hideCost` to `formatUsageDataRow`, `formatTotalsRow`, `pushBreakdownRows`
- Update all commands to conditionally hide cost column

Closes #625

## Summary by CodeRabbit

* **New Features**
  * Added new option to hide cost columns from all usage reports and data tables, including daily, monthly, session, and block analysis views. This feature provides a cleaner viewing experience and is specifically designed for Pro subscribers who prefer to focus exclusively on usage metrics without cost information being displayed throughout their reports.

### DIFF
--- a/apps/ccusage/src/_shared-args.ts
+++ b/apps/ccusage/src/_shared-args.ts
@@ -108,6 +108,11 @@ export const sharedArgs = {
 		description: 'Force compact mode for narrow displays (better for screenshots)',
 		default: false,
 	},
+	noCost: {
+		type: 'boolean',
+		description: 'Hide cost column from output (useful for Pro subscribers)',
+		default: false,
+	},
 } as const satisfies Args;
 
 /**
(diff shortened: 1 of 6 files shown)


ryoppippi:

I am sorry that the Rust migration made this no longer mergeable as-is. The production CLI has moved from the TypeScript implementation to the Rust implementation under `rust/crates/ccusage`, and this PR targets the retired TypeScript runtime path. This PR was also opened in 2025, so it is too old to merge safely against the current `main`. I am closing it.

If this idea is still needed, please reopen it as a fresh Rust implementation against the current CLI. The original idea may still be useful, but this TypeScript patch is obsolete now.

Issue #705, opened 2025-10-24
Output token count is highly inaccurate.


maxandrews:
Anthropic changed the way output tokens are treated in the logs. Neither thinking tokens nor tool use tokens are being correctly counted.

You need to use a more complex method to reconstruct accurate output tokens from the other values. 

See this example from writing a fairly long markdown file: 

<img width="258" height="23" alt="Image" src="https://github.com/user-attachments/assets/c5d898b5-ddda-4748-abc3-d7c25beb1c79" />
<img width="148" height="22" alt="Image" src="https://github.com/user-attachments/assets/4f20d5b1-a853-4ec4-97fb-83b98add557e" />

This is 8836 output tokens, not 13!

jstasiak:
Hey, I'm the author of that PR, while that patch looks like _a_ fix I'm not sure if it's _the_ fix and if it actually addresses the problem reported here.

There are sessions (or whole days) where the numbers produced by my patched version match the upstream ccusage's numbers, there are days where there are significant differences, I'm thinking I should still dig into that, I know there were much larger differences in early January in my case.

I'd appreciate if someone experiencing severe output token undercounting could test the patch in their environment and compare the numbers, this would be very valuable information.

ryoppippi:
Closing as a duplicate of #866. This older output-token accuracy report is covered by the broader current finding that Claude Code JSONL input/output usage values can be incomplete upstream, especially for thinking/tool-use output.

PR #835, opened 2026-02-01
fix(ccusage): Fix underreporting output tokens

jstasiak:
Claude Code writes multiple JSONL entries per API response during streaming. Each entry shares the same messageId:requestId hash, but output_tokens accumulates incrementally, starting near 0 and reaching the final count when the response completes.

The old dedup logic kept the first entry per hash, resulting in using low output_tokens values for some responses. The impact of that varied.

In my Claude Code sessions sometimes there was no difference at all (presumably because of no incremental response streaming), sometimes it made ~15% difference (in cost terms) in a given 5-hour block.

---

Review note: I'm not sure the change to the `should deduplicate entries across sessions` test makes sense and matches the goals of the project. I figure we'll have a problem regardless of the way we assign tokens (session 1 vs session 2)? Is there something more sophisticated we can do here?

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of streaming output token counts in usage data by ensuring the most complete entry state is retained during processing.

### DIFF
--- a/apps/ccusage/src/data-loader.ts
+++ b/apps/ccusage/src/data-loader.ts
@@ -487,25 +487,6 @@ function filterByProject<T>(
 }
 
 /**
- * Checks if an entry is a duplicate based on hash
- */
-function isDuplicateEntry(uniqueHash: string | null, processedHashes: Set<string>): boolean {
-	if (uniqueHash == null) {
-		return false;
-	}
-	return processedHashes.has(uniqueHash);
-}
-
-/**
- * Marks an entry as processed
- */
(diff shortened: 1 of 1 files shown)


jstasiak:
It's more than likely that this fix, while being an improvement, is incomplete: https://github.com/ryoppippi/ccusage/issues/705#issuecomment-3849691634

ryoppippi:
Closing during repository triage. This has been superseded by later merged work/current main, and the PR is now stale/conflicting. If a remaining gap exists, it should be reopened as a fresh narrow PR against current main.

PR #877, opened 2026-03-06
perf(codex): skip out-of-range date directories when --since/--until is set

jleechan2015:
### Root cause (Codex)

`loadTokenUsageEvents()` in `apps/codex/src/data-loader.ts` used `glob('**/*.jsonl', { cwd: sessionsDir })` — a full recursive scan unconditionally, regardless of `--since`/`--until`. The `since`/`until` values were forwarded only to the **report-building** step, not to the file-discovery step.

### Root cause (ccusage / Claude sessions)

`globUsageFiles()` in `apps/ccusage/src/data-loader.ts` globbed all `~/.claude/projects/**/*.jsonl` files (28 800+ files / 6.7 GB on large installs) before any date filtering. Claude sessions have no date directory hierarchy, so tree-traversal pruning is not applicable.

### Observed failure

On large installs the session directory exceeds available memory; macOS OOM-kills the Node process:
```
Command ccusage-codex daily --since 20260227 --order desc --json exited with code null
```

---

### Codex — directory-tree traversal pruning (`apps/codex/src/data-loader.ts`)

Added `listSessionFiles()` which enumerates `YYYY/MM/DD` directories and skips entire subtrees outside the requested date range before globbing. Falls back to `**/*.jsonl` when no date filter is given. Also preserves legacy flat `*.jsonl` files at the sessions root (reviewer feedback).

Measured on 35 GB / 12 000+ file store with `--since 7d`:
- Before: OOM kill
- After: completes in ~4 s

### ccusage — mtime-based file pruning (`apps/ccusage/src/data-loader.ts`)

Added `filterFilesByMtime()` which uses `stat()` to skip files whose mtime is more than one day before `--since`. The 24-hour buffer absorbs timezone differences. `--until` is intentionally not used for mtime pruning (a file updated after `--until` may still hold entries from before it).

Applied in `globUsageFiles()`, `loadDailyUsageData()`, `loadSessionData()`, and `loadSessionBlockData()`.

Measured on 28 800-file / 6.7 GB Claude store with `--since 7d`:
- Before: ~90 s
- After: ~25 s (10 000 files pass mtime; ~19 000 pruned at stat level)

---

## TDD methodology

RED → GREEN for each fix. In-source vitest tests added:
- Codex: `skips date directories outside the since/until range` + `includes root-level *.jsonl files when date filters are active`
- ccusage: `skips files whose mtime is more than one day before since` (uses `utimes()` to set deterministic mtime on fixture files)

---

## Impact

| Scenario | Before | After |
|---|---|---|
| `ccusage-codex --since 7d` on 35 GB store | OOM kill | ~4 s |
| `ccusage --since 7d` on 6.7 GB store | ~90 s | ~25 s |
| No date filters | Full glob (unchanged) | Full glob (unchanged) |

## Summary by CodeRabbit

* **New Features**
  * Date-range filtering for token-usage reports across daily, monthly, and session views—specify since/until to limit data.
  * Improved session/usage file selection: supports dated directory traversal and legacy root-level files, with timezone-aware mtime pruning to reduce false positives.
  * Preserves existing output shape and compatibility while enabling more targeted, efficient queries.

### DIFF
--- a/apps/ccusage/src/data-loader.ts
+++ b/apps/ccusage/src/data-loader.ts
@@ -13,7 +13,7 @@ import type { LoadedUsageEntry, SessionBlock } from './_session-blocks.ts';
 import type { ActivityDate, Bucket, CostMode, ModelName, SortOrder, Version } from './_types.ts';
 import { Buffer } from 'node:buffer';
 import { createReadStream, createWriteStream } from 'node:fs';
-import { readFile } from 'node:fs/promises';
+import { readFile, stat } from 'node:fs/promises';
 import path from 'node:path';
 import process from 'node:process';
 import { createInterface } from 'node:readline';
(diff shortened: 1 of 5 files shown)


ryoppippi:

PR #879, opened 2026-03-07
feat(opencode): migrate from JSON files to SQLite database

yukukotani:
## Summary

OpenCode has migrated data store from JSON files to SQLite database in [v1.2.0](https://github.com/anomalyco/opencode/releases/tag/v1.2.0).

This PR follows that migration.

## Note

- This PR introduces breaking changes. It will not work with OpenCode under v1.2.0.

## Summary by CodeRabbit

* **New Features**
  * Data storage migrated from per-message files to a centralized SQLite database for sessions and messages.

* **Documentation**
  * Updated docs and README to reflect the new database-backed storage model and token/cost data access.

* **Chores**
  * Bumped minimum Node.js requirement to v22.13.0.
  * Removed unused development dependencies.

### DIFF
--- a/apps/opencode/package.json
+++ b/apps/opencode/package.json
@@ -32,7 +32,7 @@
 		}
 	},
 	"engines": {
-		"node": ">=20.19.4"
+		"node": ">=22.13.0"
 	},
 	"scripts": {
 		"build": "tsdown",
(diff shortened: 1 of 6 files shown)


apps/opencode/package.json
```diff
@@ -32,7 +32,7 @@
 		}
 	},
 	"engines": {
-		"node": ">=20.19.4"
+		"node": ">=22.13.0"
```
yukukotani, line 35: Bumping this since `node:sqlite` was introduced in Node v22. We can stick with v20 if adding a dependency is okay.

cristoslc:
Your `role='assistant'` SQL filter was a good call — we adopted it in #943. We kept the legacy JSON fallback (non-breaking) since some installs may have historical data not backfilled into the DB. Would value your take on the cross-source merge approach.

ryoppippi:
Closing during repository triage. This is superseded by the current main branch implementation, and the PR is now conflicting with the newer ccusage architecture. Thanks for the work here; if there is still a remaining gap, it should come back as a fresh PR against the current unified adapter layout.

PR #850, opened 2026-02-18
fix(opencode): support SQLite database for OpenCode >= 1.2.2

dobbymaniac:
## Summary

Fixes #845

OpenCode 1.2.2+ migrated from individual JSON files (`storage/message/*.json`) to a SQLite database (`opencode.db`). This PR adds SQLite reading support with automatic fallback to legacy JSON files for older OpenCode versions.

## Changes

- **`apps/opencode/src/data-loader.ts`** — Added `openSqliteDb()` adapter that tries `better-sqlite3` then `bun:sqlite`, plus `loadMessagesFromSqlite()` and `loadSessionsFromSqlite()` functions. Public API (`loadOpenCodeMessages`, `loadOpenCodeSessions`) now tries SQLite first, falls back to JSON.
- **`apps/opencode/package.json`** — Added `better-sqlite3` as a runtime dependency.
- **`apps/opencode/tsdown.config.ts`** — Marked `better-sqlite3` as external (native addon, cannot be bundled).
- **`pnpm-workspace.yaml`** — Added `better-sqlite3` to catalog and `allowBuilds`.

## How it works

1. Check if `opencode.db` exists in the OpenCode data directory
2. If yes → query `message` and `session` tables via SQLite
3. If no (or SQLite fails) → fall back to legacy JSON file globbing

The SQLite `message.data` column contains the same JSON structure as the old files, so the parsing/filtering logic is equivalent.

## Testing

- All 6 existing tests pass
- Verified `daily`, `session`, and `weekly` commands against a real OpenCode 1.2.2+ database

## Release Notes

* **New Features**
  * OpenCode now supports efficient local database loading for improved performance, with automatic fallback to existing formats.
  * Extended token tracking to include reasoning data alongside input and output metrics.

* **Chores**
  * Updated build and workspace configuration for optimized dependency management.

PR #887, opened 2026-03-10
feature(opencode): support opencode sqlite storage

unsync:
## Summary

Initial credits to https://github.com/ryoppippi/ccusage/pull/850
Duplicate: https://github.com/ryoppippi/ccusage/pull/879

### Why this change was made
OpenCode moved its usage data to `opencode.db` in newer releases, so `ccusage` was missing recent OpenCode activity and falling back to stale legacy JSON data.

### Description
This PR adds SQLite-backed OpenCode loading with a legacy JSON fallback so the OpenCode reports stay accurate across both old and new OpenCode storage formats.
- Reads OpenCode usage and session metadata from `~/.local/share/opencode/opencode.db` when present, while preserving the existing JSON-file path for older installs.
- Keeps `daily` and `session` reports consistent by sharing one combined SQLite load path instead of re-reading usage and session data separately.
- Falls back to `bun:sqlite` when `better-sqlite3` is unavailable or cannot open under Bun, which fixes local Bun-based CLI runs.
- Adds focused loader tests for SQLite parsing, deduplication, timestamp fallback, and session metadata defaults.
- Adds `better-sqlite3` as an optional OpenCode dependency and updates workspace build approval/catalog entries for the new native package.

### Blast Radius
- Affects only the OpenCode CLI package under `apps/opencode` plus the minimal workspace dependency metadata needed to ship the SQLite reader.
- Impacts users running `ccusage` against newer OpenCode installs that now store usage in SQLite instead of JSON files.

### Practical Consequence
- After deploy, OpenCode reports should include recent usage again instead of stopping at the last legacy JSON records.
- Bun-based local runs continue to work even though Bun cannot use `better-sqlite3` directly, because the loader now falls back to `bun:sqlite` instead of dropping to legacy JSON.

### Decisions
- Kept the change scoped to OpenCode data loading and package metadata instead of refactoring unrelated report logic.
- Made `better-sqlite3` optional and externalized it so Node builds can use it while Bun runtime still has a supported fallback.
- Preserved the legacy JSON loader as a compatibility path for older OpenCode versions and as a recovery path when SQLite reads fail.

### Verification
- Ran `pnpm --filter @ccusage/opencode run test` successfully.
- Ran `pnpm --filter ccusage run test` successfully.
- Ran `pnpm typecheck` successfully.
- Ran `pnpm --filter @ccusage/opencode run start daily` successfully under Bun and confirmed it now prints a real report instead of falling back after a SQLite open failure.

## Summary by CodeRabbit

* **New Features**
  * Support for faster OpenCode data loading from local SQLite databases (OpenCode ≥ 1.2.2) with automatic fallback to legacy JSON for compatibility.

* **Chores**
  * Enabled optional SQLite runtime support and updated build config to allow bundling the SQLite runtime when present.

* **Tests**
  * Added tests and test helpers covering SQLite loading, deduplication, filtering, timestamp handling, and session metadata.

### DIFF
--- a/apps/opencode/src/cost-utils.ts
+++ b/apps/opencode/src/cost-utils.ts
@@ -11,8 +11,13 @@ const MODEL_ALIASES: Record<string, string> = {
 	'gemini-3-pro-high': 'gemini-3-pro-preview',
 };
 
+function normalizeClaudeModelName(modelName: string): string {
+	return modelName.replace(/^claude-(opus|sonnet|haiku)-(\d+)\.(\d+)$/, 'claude-$1-$2-$3');
+}
+
 function resolveModelName(modelName: string): string {
-	return MODEL_ALIASES[modelName] ?? modelName;
+	const normalizedModelName = normalizeClaudeModelName(modelName);
+	return MODEL_ALIASES[normalizedModelName] ?? normalizedModelName;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


yukukotani:
@unsync Thanks for sharing that! It seems like just a trade-off between simplicity and compatibility. #879 is simple and less dependencies but not compatible for older opencode. And I believe @ryoppippi loves simplicity.

unsync:
> @unsync Thanks for sharing that! It seems like just a trade-off between simplicity and compatibility. #879 is simple and less dependencies but not compatible for older opencode. And I believe @ryoppippi loves simplicity.

@yukukotani thanks for the reply !

If it was up to me alone, your PR is fine, only two points i see is:
- the node 22 requirement which is more than an opencode-only compatibility
- the reported bug on claude models usage that I fixed here

If @ryoppippi favor your PR I'd be glad to add my usage fix on your PR so we don't need to create another one 🙏

entrptaher:
Wondering why this was not merged. For now I have to clone the repo manually and test it there.

cristoslc:
Your `SqliteAdapter` abstraction and `bun:sqlite` fallback were the right approach — we adopted both in #943 along with your better-sqlite3-as-external pattern. Our PR also adds cross-source merge (DB + legacy files), channel-variant DB detection, and path caching. Note: we scoped out your `normalizeClaudeModelName()` fix as a separate PR since it's a pricing lookup concern, not a SQLite concern. Would value your review.

unsync:
@cristoslc followed your progress, but didn't dive into the code, will try.

I'm closing in favor of https://github.com/ryoppippi/ccusage/pull/943 or https://github.com/ryoppippi/ccusage/pull/879 from @yukukotani to avoid the PR bloat around this.

PR #898, opened 2026-03-18
fix(ccusage): add flat pricing for Opus 4.6 / Sonnet 4.6 to fix inflated costs

### LINKED ISSUE #891: Inflated cost calculation since Opus 1M context enabled in 2.1.75
I think 1M context was enabled on Opus in claude code version 2.1.75

Since then ccusage shows way higher cost. The input/output tokens on 14th March are lower than on 12th march but still the cost is more than 5x. Cache read is around 4x higher on 14th March but that can't account for the cost.

Image

MumuTW:
## Summary

- Fixes inflated cost calculation for Claude Opus 4.6 since Claude Code enabled 1M context in v2.1.75
- Claude Opus 4.6 and Sonnet 4.6 use **flat pricing** across the full 1M context window — no premium for requests exceeding 200k input tokens ([docs](https://docs.anthropic.com/en/docs/about-claude/pricing#long-context-pricing))
- LiteLLM's dataset only has Bedrock entries (`anthropic.claude-opus-4-6-v1`) which include `*_above_200k_tokens` tiered pricing fields; when ccusage falls back to substring matching it picks up the Bedrock entry and incorrectly applies tiered pricing
- Fix: inject entries keyed by the exact Anthropic API model names (`claude-opus-4-6`, `claude-sonnet-4-6`) with correct flat pricing so direct lookup succeeds before the substring fallback fires

## Test plan

- [x] All 252 existing ccusage tests pass
- [x] All 9 internal pricing tests pass
- [ ] Verify with a real Claude Code session that Opus 4.6 costs are no longer inflated

Fixes #891

## Summary by CodeRabbit

* **New Features**
  * Added support for injecting first-party flat pricing entries for specific Anthropic models so direct API lookups use precise flat rates.
* **Bug Fixes**
  * Improved cost calculation accuracy for Claude Opus/Sonnet models when using direct Anthropic pricing.
* **Tests**
  * Added tests verifying flat-pricing calculations and override behavior.

### DIFF
--- a/apps/ccusage/src/_macro.ts
+++ b/apps/ccusage/src/_macro.ts
@@ -13,10 +13,47 @@ function isClaudeModel(modelName: string, _pricing: LiteLLMModelPricing): boolea
 	);
 }
 
+/**
+ * Anthropic API pricing for 1M-context models that use flat (non-tiered) pricing.
+ *
+ * Claude Opus 4.6 and Sonnet 4.6 include the full 1M context window at standard
+ * pricing — there is no premium for requests exceeding 200k input tokens.
+ * See: https://docs.anthropic.com/en/docs/about-claude/pricing#long-context-pricing
+ *
+ * LiteLLM's dataset only contains Bedrock entries (e.g. `anthropic.claude-opus-4-6-v1`)
+ * which carry `*_above_200k_tokens` fields for Bedrock-specific tiered pricing.
+ * When ccusage falls back to substring matching, it picks up those Bedrock entries
+ * and incorrectly applies tiered pricing to direct Anthropic API usage, inflating costs.
+ *
+ * By injecting entries keyed by the exact Anthropic API model names, direct lookup
+ * succeeds before the substring fallback ever fires.
+ */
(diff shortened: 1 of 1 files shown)


ryoppippi:
Thank you for opening this and for digging into the inflated Opus 4.6 pricing issue.

After checking the current Anthropic pricing docs and the latest LiteLLM pricing dataset, I think we should not merge this as-is. LiteLLM now has first-party claude-opus-4-6, claude-opus-4-7, and claude-sonnet-4-6 entries with flat 1M-context pricing, so ccusage should avoid hardcoding those prices locally.

Also, this does not apply to every 4.5 model: claude-sonnet-4-5 still has above-200k tier fields in LiteLLM, while the 1M flat long-context pricing is currently documented for Opus 4.7, Opus 4.6, and Sonnet 4.6.

I will close this in favour of a narrower PR that relies on LiteLLM first-party flat entries and avoids falling through to provider-specific substring matches where that would select the wrong pricing. Sorry for the churn here, and thanks again for catching the underlying issue.

ryoppippi:
Closing this in favour of a narrower follow-up PR that relies on LiteLLM first-party flat pricing entries instead of hardcoding model prices in ccusage.

PR #910, opened 2026-03-27
[codex] fix forked session double-counting and optimize loading

jackcpku:
## Why this PR matters

- Fixes a correctness bug: forked Codex sessions can replay an ancestor `token_count` prefix, and the previous loader counted that replay again. That inflated usage totals for fork-heavy workflows such as subagents.
- Makes large Codex histories much more practical to analyze by switching session loading to a streaming path with earlier filtering and online aggregation.

## Headline impact

| Area | Before | After |
| --- | --- | --- |
| Forked session accounting | replayed ancestor usage could be counted again | replayed ancestor prefixes are ignored |
| Large-history reliability | `daily` could appear hung and use multi-GB RSS | queries complete reliably with much lower peak memory |
| Narrow date filters | full-history scan first, filter later | file overlap is checked before parsing |

## Benchmark impact

> Summary: on the same one-day subset benchmark, this PR cuts peak RSS by about 55% at the cost of higher wall time. The main win is correctness for forked sessions plus avoiding multi-GB behavior on larger histories.

| Scenario | Before | After | Improvement |
| --- | ---: | ---: | ---: |
| Exact benchmark requested: `CODEX_HOME=/Users/jack/tmp/codex-home-one-day node dist/index.js daily --offline --since 20260327 --until 20260327 >/dev/null` | baseline `1.82s real`, `1243414528` max RSS | optimized `3.04s real`, `554549248` max RSS | about `55%` lower peak RSS, but wall time is slower on this tiny subset |
| Real local history: `daily --offline --since 20260327 --until 20260327` | effectively hung / very slow in earlier reproduction | `9.75s real`, `695648256` max RSS | command now completes reliably with much lower memory |
| Synthetic large-file prototype comparison | current package: `5.02s`, `5.42 GB` peak | optimized prototype: `1.15s`, `152 MB` peak | about `4.4x` faster, about `36x` less memory |
| Narrow date-range scan potential on this dataset | full scan: about `46G` | one-day directory: about `224 MB` | about `220x` less data to read when filters are narrow |

## Root cause

The previous implementation loaded `~/.codex/sessions/**/*.jsonl` with `readFile`, split the whole file into lines, validated/parsing each line in the hot path, pushed every event into arrays, and only then applied reporting filters and aggregation.

It also ignored `session_meta.payload.forked_from_id`, so when a child session replayed ancestor `token_count` history the loader treated that replay as fresh usage and double-counted it.

## What changed

- switched `apps/codex/src/data-loader.ts` from `readFile + split + per-line schema parsing` to streaming line-by-line parsing
- added file-level date overlap filtering before reading session files
- added callback-based loading so commands can aggregate online without materializing all events
- updated `daily`, `monthly`, and `session` commands to build summaries directly during load instead of collecting and sorting every event first
- kept the existing report builders, but refactored them so they can also finalize from pre-aggregated summaries
- deduplicated replayed `token_count` prefixes in forked sessions by following `forked_from_id` ancestry during load
- added coverage for streaming callback mode, date-range file skipping, and forked-session replay deduplication

## Validation

Checks run:

- `pnpm --filter @ccusage/codex test -- src/data-loader.ts`
- `pnpm --filter @ccusage/codex lint`
- `pnpm --filter @ccusage/codex typecheck`
- `pnpm --filter @ccusage/codex build`

CLI validation:

- baseline benchmark in a separate worktree: `cd /Users/jack/tmp/ccusage_repo-baseline/apps/codex && CODEX_HOME=/Users/jack/tmp/codex-home-one-day /usr/bin/time -l node dist/index.js daily --offline --since 20260327 --until 20260327 >/dev/null`
- `node dist/index.js daily --offline --since 20260327 --until 20260327` on the real local Codex history completed in `9.75s real` with `695648256` max RSS
- `CODEX_HOME=/Users/jack/tmp/codex-home-one-day node dist/index.js daily --offline --since 20260327 --until 20260327` completed in `3.04s real` with `554549248` max RSS
- verified `session` output on the real local history for `2026-03-27`
- verified `monthly` output on a one-day Codex subset home
- added regression coverage that deduplicates replayed fork prefixes and still works when the ancestor file falls outside the requested date range

### DIFF
--- a/apps/codex/src/commands/daily.ts
+++ b/apps/codex/src/commands/daily.ts
@@ -1,3 +1,4 @@
+import type { DailyUsageSummary } from '../_types.ts';
 import process from 'node:process';
 import {
 	addEmptySeparatorRow,
@@ -12,7 +13,7 @@ import pc from 'picocolors';
 import { DEFAULT_TIMEZONE } from '../_consts.ts';
 import { sharedArgs } from '../_shared-args.ts';
 import { formatModelsList, splitUsageTokens } from '../command-utils.ts';
-import { buildDailyReport } from '../daily-report.ts';
+import { accumulateDailyUsage, buildDailyReportRows } from '../daily-report.ts';
 import { loadTokenUsageEvents } from '../data-loader.ts';
 import { normalizeFilterDate } from '../date-utils.ts';
 import { log, logger } from '../logger.ts';
@@ -41,13 +42,23 @@ export const dailyCommand = define({
 			process.exit(1);
 		}
 
-		const { events, missingDirectories } = await loadTokenUsageEvents();
+		const summaries = new Map<string, DailyUsageSummary>();
+		const { missingDirectories } = await loadTokenUsageEvents({
+			since,
+			until,
+			timezone: ctx.values.timezone,
+			collectEvents: false,
+			sortEvents: false,
+			onEvent: (event) => {
+				accumulateDailyUsage(summaries, event, ctx.values.timezone);
+			},
+		});
 
 		for (const missing of missingDirectories) {
 			logger.warn(`Codex session directory not found: ${missing}`);
 		}
 
-		if (events.length === 0) {
+		if (summaries.size === 0) {
 			log(jsonOutput ? JSON.stringify({ daily: [], totals: null }) : 'No Codex usage data found.');
 			return;
 		}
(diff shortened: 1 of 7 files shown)


MaxGhenis:
I reviewed this against a real Codex corpus and found two additional correctness issues in the new loader path:

- some real logs emit a nonzero `last_token_usage` while `total_token_usage` is unchanged, so preferring `last_token_usage` can still overcount badly
- mixed `last_token_usage`-only then `total_token_usage` records double count unless the running baseline advances on the `last`-only events

I sent a small follow-up PR against the current branch here: https://github.com/jackcpku/ccusage/pull/1

It also lets `onEvent` callback errors bubble out instead of being swallowed as stream-read failures.

ryoppippi:
Closing during repository triage. This is superseded by current main or by later merged work, and it conflicts with the current unified adapter architecture. If a gap remains, it needs a fresh narrow PR against main.

PR #923, opened 2026-03-31
fix(statusline): use current_usage instead of total_input_tokens for context display

AndriiBarabash:
I'm sorry to open this pull request without an attached issue to it. My organization recently switched to Claude Code and I installed your project to track my token usage. The regular `ccusage` works as expected, but then I installed it to the Claude Code status line (via `ccusage statusline`) and saw the incorrect token calculations near the 🧠 emoji. 

So here is the PR with the fix for this issue. After the fix, the tokens in statusline match the tokens with `/config -> Verbose output (set to true)`.

## Summary

This fix resolves an issue where the statusline command (`ccusage statusline`) was displaying incorrect context window token counts and percentages.

See the testing section for before/after.

## The Problem

The statusline was reading `hookData.context_window.total_input_tokens`, which represents the **cumulative session input tokens** rather than the actual tokens currently in the context window. This led to wildly inaccurate displays.

## The Solution

Now calculates actual context window usage by summing three components from `current_usage`:
- `input_tokens` - current input in context
- `cache_creation_input_tokens` - tokens written to cache
- `cache_read_input_tokens` - tokens read from cache

This matches how Claude Code itself calculates and displays context usage, providing users with accurate real-time visibility.

## Additional Changes

- Updated `statuslineHookJsonSchema` to properly validate the `current_usage`, `used_percentage`, and `remaining_percentage` fields from Claude Code's statusline hook
- Made these fields nullable to handle fresh sessions where no API calls have been made yet (`current_usage` is `null`)
- When `current_usage` is unavailable but context window size is known, displays `0 (0%)` instead of falling back to transcript calculation or showing "N/A"

## Testing

Verified that statusline output now matches Claude Code's context display:
- Before:
Screenshot 2026-03-31 at 23 08 32

- After:
Screenshot 2026-03-31 at 23 07 44

## Summary by CodeRabbit

* **Bug Fixes**
  * Status display now prefers explicit current-usage metrics when available, summing input and cached tokens for more accurate context counts. If transcript-based estimation fails but a context window size is present, context tokens are shown as 0 instead of undefined.
* **New Features**
  * Added optional current-usage, used-percentage, and remaining-percentage fields to improve status reporting fidelity and clarity.

### DIFF
--- a/apps/ccusage/src/_types.ts
+++ b/apps/ccusage/src/_types.ts
@@ -184,6 +184,16 @@ export const statuslineHookJsonSchema = v.object({
 			total_input_tokens: v.number(),
 			total_output_tokens: v.optional(v.number()),
 			context_window_size: v.number(),
+			current_usage: v.optional(
+				v.object({
+					input_tokens: v.number(),
+					output_tokens: v.number(),
+					cache_creation_input_tokens: v.number(),
+					cache_read_input_tokens: v.number(),
+				}),
+			),
+			used_percentage: v.optional(v.number()),
+			remaining_percentage: v.optional(v.number()),
 		}),
 	),
 });
--- a/apps/ccusage/src/_types.ts
+++ b/apps/ccusage/src/_types.ts
@@ -184,7 +184,7 @@ export const statuslineHookJsonSchema = v.object({
 			total_input_tokens: v.number(),
 			total_output_tokens: v.optional(v.number()),
 			context_window_size: v.number(),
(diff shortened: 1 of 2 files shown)


AndriiBarabash:
@ryoppippi the PR is ready for review

AndriiBarabash:
@coderabbitai resume

AndriiBarabash:
@coderabbitai review

ryoppippi:

PR #940, opened 2026-04-15
feat(ccusage): add --human/-H flag for human-readable token display

jzlikewei:
## Summary

- Add `--human` / `-H` flag to all commands (daily, weekly, monthly, session, blocks)
- When enabled, token counts are displayed with K/M/B suffixes (e.g., `1.50M` instead of `1,500,000`)
- Includes unit tests for `formatNumberHuman` and `setHumanReadableNumbers`

## Details

Large token counts (millions+) are hard to scan visually. The `--human` flag provides a compact format:

| Raw | Human |
|---|---|
| 1,500,000 | 1.50M |
| 2,500 | 2.50K |
| 999 | 999 |

Implementation:
- New `formatNumberHuman()` function in `packages/terminal/src/table.ts`
- Global toggle via `setHumanReadableNumbers()` — called before any table output when `--human` is passed
- `formatNumber()` automatically delegates to `formatNumberHuman()` when the flag is active

## Summary by CodeRabbit

* **New Features**
  * Added `-H` / `--human` CLI flag to show token counts in human-readable format (K/M/B suffixes, rounded) in non-JSON tables and reports across blocks, daily, weekly, monthly, and session commands. JSON output remains unchanged.

### DIFF
--- a/apps/ccusage/src/_shared-args.ts
+++ b/apps/ccusage/src/_shared-args.ts
@@ -110,6 +110,12 @@ export const sharedArgs = {
 		description: 'Force compact mode for narrow displays (better for screenshots)',
 		default: false,
 	},
+	human: {
+		type: 'boolean',
+		short: 'H',
+		description: 'Display token counts in human-readable format (K/M/B suffixes)',
+		default: false,
+	},
 } as const satisfies Args;
 
 /**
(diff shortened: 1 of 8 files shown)


ryoppippi:

PR #942, opened 2026-04-15
fix(opencode): add SQLite data loader for new OpenCode storage backend

cristoslc:
## Summary

OpenCode recently switched from storing messages as individual JSON files (`storage/message/*.json`) to a SQLite database (`opencode.db`). The existing data loader in this package only knew the file-based format, so users on newer OpenCode versions would see no usage data at all.

- Adds SQLite loading via `bun:sqlite` (preferred path when `opencode.db` is detected in the OpenCode data directory)
- Keeps the JSON file-based path as a fallback for older OpenCode installations
- Adds a thin Node.js shim (`_sqlite-node.ts`) that re-exports `better-sqlite3` under the `bun:sqlite` module ID, with alias config in both `tsdown` and `vitest` for cross-runtime compatibility (Bun at runtime, Node.js in CI/tests)
- Adds `better-sqlite3` / `@types/better-sqlite3` as dev dependencies and registers `better-sqlite3` in the pnpm workspace catalog with `allowBuilds`
- Includes in-source vitest tests covering both the SQLite path and the legacy JSON file path

## Cross-runtime compatibility approach

The OpenCode CLI is a Bun-native binary, so `bun:sqlite` is the natural import. However, vitest runs under Node.js. The shim + alias pattern means:

- **Bun (production):** `bun:sqlite` resolves to Bun's built-in SQLite module directly
- **Node.js / vitest:** `bun:sqlite` is aliased to `apps/opencode/src/_sqlite-node.ts`, which re-exports `better-sqlite3` with a compatible `Database` API

## Test plan

- [ ] Run `pnpm run test` — existing and new OpenCode tests pass under Node.js/vitest
- [ ] Verify `pnpm run build` succeeds for the opencode package
- [ ] On a machine with a newer OpenCode installation (has `opencode.db`), confirm usage data is loaded and displayed correctly
- [ ] On a machine with an older OpenCode installation (only JSON files), confirm the fallback path still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cristoslc:
Closing to refactor — the bun:sqlite abstraction layer adds unnecessary complexity for upstream. Will reopen as a simpler PR using better-sqlite3 directly, which is more conventional and keeps the diff smaller.

PR #970, opened 2026-05-02
fix(statusline): use configured timezone for today cost

### LINKED ISSUE #778: Statusline "today" shows $0.00 due to timezone defaulting to UTC
When using ccusage statusline, the "today" value shows $0.00 even though ccusage daily correctly shows usage for the current day. This appears to be a timezone issue - the statusline seems to default to UTC rather than detecting the local timezone.

### Environment:
ccusage: 16.1.2
Runtime: Bun 1.3.5
OS: Ubuntu (WSL2)
Timezone: Australia/Perth (UTC+8)

### Steps to reproduce:
Be in a timezone significantly offset from UTC (e.g., UTC+8)
Use Claude Code during local daytime hours
Configure statusline in `~/.claude/settings.json`
Observe "today" shows $0.00 while "session" correctly increments
Run `bunx ccusage daily` - shows correct usage for the day

### Workaround:
Creating a `ccusage.json` config file in the `~/.claude.` directory with the timezone setting resolves the issue.
```
{
  "$schema": "https://ccusage.com/config-schema.json",
  "defaults": {
    "timezone": "Australia/Perth"
  }
}
```

### Suggestion:
Consider having the statusline default to the system's local timezone (like daily appears to do) rather than UTC, or document that timezone configuration is required for accurate "today" aggregation.

aukaukauk:
Fixes #778.

## Summary

`statusline` computed today's date with `Date.toISOString()`, which always uses UTC. Daily usage is grouped using the local or configured timezone, so users far from UTC could see `$0.00 today` during date-boundary windows.

This changes statusline to compute the today filter with the same timezone-aware formatter used by daily grouping, passes that timezone into `loadDailyUsageData`, and exposes `--timezone` on the statusline command.

## Verification

- `pnpm --filter ccusage run format`
- `pnpm --filter ccusage typecheck`
- `pnpm --filter ccusage run test`
- `git diff --check origin/main...HEAD`

## Release Notes

* **New Features**
  * Added timezone configuration setting to statusline. This enables accurate date grouping and filtering based on your specified timezone instead of the system default, ensuring proper daily usage tracking across different regions.

### DIFF
--- a/apps/ccusage/config-schema.json
+++ b/apps/ccusage/config-schema.json
@@ -665,6 +665,11 @@
 									"markdownDescription": "Context usage percentage below which status is shown in yellow (0-100)",
 									"default": 80
 								},
+								"timezone": {
+									"type": "string",
+									"description": "Timezone for date grouping (e.g., UTC, America/New_York, Asia/Tokyo). Default: system timezone",
+									"markdownDescription": "Timezone for date grouping (e.g., UTC, America/New_York, Asia/Tokyo). Default: system timezone"
+								},
 								"debug": {
 									"type": "boolean",
 									"description": "Show pricing mismatch information for debugging",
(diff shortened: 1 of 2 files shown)


ryoppippi:

PR #980, opened 2026-05-11
fix(ccusage): avoid provider-specific Claude pricing fallback

ryoppippi:
## Summary

- Restrict substring fallback for first-party Claude model names to first-party LiteLLM entries.
- Avoid selecting provider-specific Bedrock/Vertex entries for first-party Claude lookups when direct LiteLLM flat entries should be used instead.
- Keep pricing values sourced from LiteLLM rather than hardcoding Opus/Sonnet prices in ccusage.

## Context

This replaces #898 with a narrower fix. Current Anthropic docs and LiteLLM data show 1M flat pricing for Opus 4.7, Opus 4.6, and Sonnet 4.6, but not every 4.5 model is flat. In particular, LiteLLM still has above-200k tier fields for Sonnet 4.5.

## Testing

- pnpm run format
- pnpm typecheck
- pnpm run test

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pricing lookup for Claude models to ensure accurate pricing information by properly distinguishing first-party Claude models from provider-specific variants.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ryoppippi/ccusage/pull/980)

### DIFF
--- a/packages/internal/src/pricing.ts
+++ b/packages/internal/src/pricing.ts
@@ -92,6 +92,10 @@ function createLogger(logger?: PricingLogger): PricingLogger {
 	};
 }
 
+function isFirstPartyClaudeModel(modelName: string): boolean {
+	return modelName.toLowerCase().startsWith('claude-');
+}
+
 export class LiteLLMPricingFetcher implements Disposable {
 	private cachedPricing: Map<string, LiteLLMModelPricing> | null = null;
 	private readonly logger: PricingLogger;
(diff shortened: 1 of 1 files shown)


ryoppippi:
Closing this for now. Current LiteLLM pricing data already includes first-party flat entries for the affected Claude 1M-context models, so carrying extra matching logic in ccusage is not worth the added implementation surface.

PR #984, opened 2026-05-12
perf(ccusage): optimize bundled cli performance

ryoppippi:
## Try This PR

The `ccusage` package should be installable through pkg.pr.new once the preview package workflow has published this PR.

```sh
pnpm dlx https://pkg.pr.new/ryoppippi/ccusage@984 -- --offline
```

Do not use this form; it does not work for this preview package:

```sh
bun x https://pkg.pr.new/ryoppippi/ccusage@984
```

With Bun available on `PATH`, the published launcher starts the bundled Bun entrypoint. Set `CCUSAGE_BUN_AUTO_RUN=0` to force the Node path.

## Summary

This PR optimizes the bundled JavaScript ccusage CLI while keeping output behavior aligned with main. The main changes are faster JSONL discovery/loading, worker-thread parsing for built dist usage, smaller worker payloads, faster aggregation/table hot paths, Bun-aware launcher support, CLI-only package exports, and CI fixture performance reporting.

The latest pass fixes a Bun stdout edge case: large final JSON/table writes now wait for the stream write callback. Without that, piping a large report through tools such as `jq`, `cat`, or `tee` could truncate stdout at 64 KiB while direct terminal output looked fine. CI preview E2E now runs `daily --offline --json | jq -e .` on the Node-forced path and the Bun-available path, and the fixture is large enough to catch this class of truncation.

## Current Benchmarks

Environment: Apple M3 Pro macOS, Bun built entry, real local Claude data, `LOG_LEVEL=0`, `COLUMNS=200`, hyperfine `--warmup 4 --runs 10`. These local numbers are directional; CI fixture compare is the repeatable PR signal.

| Command | Current JS/Bun |
| --- | ---: |
| `daily --offline --json` | `329.4ms ± 5.2ms` |
| `session --offline --json` | `342.5ms ± 13.3ms` |
| `blocks --offline --json` | `387.3ms ± 7.0ms` |
| `daily --offline` table | `354.4ms ± 22.1ms` |

Latest commit `213cb95` replaces the hot string-key dedupe `Map`s with null-prototype object indexes after CPU profiling showed `Map#set` dominating the post-worker merge. Compared with the post-stdout-drain run, this improved daily `336.7ms -> 329.4ms`, session `346.2ms -> 342.5ms`, and blocks `389.7ms -> 387.3ms`.

Minify A/B on valid `/dist/index.js` paths before the stdout-drain fix:

| Command | Non-minified | Minified |
| --- | ---: | ---: |
| `daily --offline --json` | `344.2ms ± 11.6ms` | `342.0ms ± 8.8ms` |
| `session --offline --json` | `368.4ms ± 23.5ms` | `351.0ms ± 10.3ms` |
| `blocks --offline --json` | `397.9ms ± 16.9ms` | `389.8ms ± 5.4ms` |

## Main And Rust References

Latest CI fixture compare against the base branch, using built `apps/ccusage/dist/index.js` through `bun -b`, `--offline --json`, mitata with `2` warmups and `7` samples:

| Command | Base median | PR median | PR vs base |
| --- | ---: | ---: | ---: |
| `daily --offline --json` | `366.8ms` | `335.6ms` | `1.09x` faster |
| `session --offline --json` | `365.3ms` | `330.9ms` | `1.10x` faster |
| `blocks --offline --json` | `357.0ms` | `333.7ms` | `1.07x` faster |

Rust reference: [#977](https://github.com/ryoppippi/ccusage/pull/977). Latest local Rust measurement on branch `rust` at `f8d6cd3`, using `target/release/ccusage` (`937 KiB`) with `LOG_LEVEL=0`, `--offline --json`, hyperfine `--warmup 4 --runs 10`: daily `249.3ms ± 6.1ms`, session `255.4ms ± 7.0ms`, blocks `260.7ms ± 9.0ms`. JS/Bun is now close enough to be practical, but this Rust binary is still faster across daily/session/blocks.

## Package Size

Current build output writes sourcemaps and reports about `1.08 MB` on disk. Runtime JS is about `124.6 kB`. Latest CI packed package comparison: base `133.73 KiB`, PR `43.26 KiB`, `90.47 KiB` smaller (`3.09x` smaller). Sourcemaps are excluded from the package.

## Correctness

Checked output parity for JSON output across `daily`, `session`, `monthly`, `weekly`, and `blocks` between minified and non-minified builds; `blocks[].projection` is time-dependent and was removed for that comparison. Representative table output matched for `daily` and `session`.

Large-output pipe checks after the stdout-drain fix:

| Command | Result |
| --- | ---: |
| `daily --offline --json | jq -e .` | valid JSON, `95,214` bytes |
| `blocks --offline --json | jq -e .` | valid JSON, `198,637` bytes |
| enlarged preview E2E `daily --offline --json | jq -e .` | valid JSON, about `349,660` bytes locally |

Preview E2E passed on `ubuntu-slim`, `macos-latest`, and `windows-latest`. The Windows log shows both the Node-forced run and the Bun-available run executing `npx --yes "$CCUSAGE_PREVIEW_URL" daily --offline --json | jq -e .`, followed by visible table output with the expected `Total` row.

Known note from earlier Rust/JS comparisons: Rust and JS previously differed in some session row counts on local comparisons while token totals matched. Any remaining semantic differences should stay documented before merge.

## Validation

Recent validation:

- `pnpm run format`
- `pnpm typecheck`
- `pnpm run test` (`32` files, `376` passed, `1` skipped)
- `pnpm --filter ccusage build`
- Built Bun `daily --offline --json | jq -e .`
- Built Bun `blocks --offline --json | jq -e .`
- Enlarged preview E2E fixture local check with `daily --offline --json | jq -e .`
- CI preview E2E on Linux, macOS, and Windows

## Latest Update (May 15, 2026, worker-side dedupe experiment)

Pushed empty commit `e29d228` to record a rejected worker-side chunk dedupe prototype. The prototype removed duplicate daily/session rows inside each worker before `postMessage`, carried source-order metadata for replacement tie-breaks, and tried to restore first-occurrence merge order before final aggregation.

I did not adopt it. JSON output was not byte-identical: token totals stayed stable, but cost fields moved at the IEEE754 tail because aggregation order changed. An intermediate shape also changed session totals because session/project metadata is encoded at the file payload level, so moving a winning row into the first-seen file slot is not semantics-preserving.

Built Bun benchmark on real local Claude data, `LOG_LEVEL=0`, `COLUMNS=200`, hyperfine `--warmup 4 --runs 10`:

| Command | Worker-dedupe prototype | Current adopted build | Result |
| --- | ---: | ---: | --- |
| `daily --offline --json` | `363.1ms ± 13.5ms` | `329.4ms ± 5.2ms` | slower |
| `session --offline --json` | `362.3ms ± 12.8ms` | `342.5ms ± 13.3ms` | slower |
| `blocks --offline --json` | `411.9ms ± 25.2ms` | `387.3ms ± 7.0ms` | slower |

After reverting the prototype and rebuilding ccusage, `daily`, `session`, and `blocks` JSON output matched the saved pre-experiment baseline byte-for-byte.

## Latest Update (May 15, 2026, object model summary indexes)

Pushed `4864bf2` for #984. Summary aggregation now uses null-prototype object indexes plus an insertion-order model list instead of allocating `Map`/`Set` for every daily/session/bucket group. This follows the same exact-string-key pattern that helped the row dedupe path, while preserving `modelsUsed` first-seen order and model breakdown sorting.

Correctness check: built Bun dist matched the pre-change output byte-for-byte for `daily`, `session`, `monthly`, `weekly`, and `blocks --offline --json`; `daily` and `session` table output were also checked and matched byte-for-byte. Added a regression test for the subtle unknown-model behavior: an undefined model contributes an `unknown` model breakdown, but `modelsUsed` only includes `unknown` when that model name is explicit.

Worker-enabled A/B, real local Claude data, `LOG_LEVEL=0`, `COLUMNS=200`, hyperfine `--warmup 4 --runs 12` with both builds under `/tmp/*/dist/index.js`:

| Command | Base | Object indexes | Result |
| --- | ---: | ---: | --- |
| `daily --offline --json` | `348.9ms ± 12.7ms` | `349.5ms ± 20.0ms` | neutral |
| `session --offline --json` | `354.1ms ± 11.8ms` | `348.5ms ± 4.0ms` | ~`1.02x` faster |
| `blocks --offline --json` | `400.9ms ± 19.9ms` | `396.0ms ± 11.3ms` | ~`1.01x` faster |

Validation: `pnpm run format`, `pnpm typecheck`, `pnpm run test` (`32` files, `377` passed, `1` skipped), `pnpm --filter ccusage build`, and built Bun `daily`/`blocks --offline --json | jq -e .`.

## Summary by CodeRabbit

* **New Features**
  * Automatic Bun re-run when available in PATH for faster warm startup (disable with `CCUSAGE_BUN_AUTO_RUN=0`)
  * Added `--singleThread` flag to disable parallel JSONL file loading

* **Bug Fixes**
  * Removed `--jq` CLI option; use piped `jq` with `--json` output instead

* **Documentation**
  * Removed library usage documentation; package is now CLI-only
  * Updated guides for new Bun auto-run behavior and JSON piping approach


erictheswift:
Real-world data point on a different corpus: this PR turns the `statusline` subcommand from unusable to instant.

**Setup**
- macOS, Claude Code statusline hook invoked per ANSI repaint (~every few seconds)
- One active conversation with a 6+ hour transcript (~tens of MB JSONL)
- Comparing `bunx ccusage@18.0.10 statusline --no-offline --cache --cost-source=both` vs. node-direct on the #984 tarball, same flags

**Numbers**

| Mode | Cold start | Warm | Resident memory |
|---|---|---|---|
| v18.0.10 (via bunx) | **~211 s** | ~1.8 s | up to ~4 GB during parse |
| #984 (direct `node dist/cli.js`) | **~500 ms** | **~140 ms** | ~50 MB transient |

Cold-start ratio ≈ **400×**.

**Side effect this unblocks**

With v18.0.10, statusline hooks that retry on stale-cache miss can stack multiple long-running ccusage processes. I had to write a lockfile + pgrep guard + process-group SIGKILL wrapper in the statusline script to prevent runaway CPU/RAM. With #984 the wrapper collapses to a synchronous `spawnSync` (28 lines, no background work, no cache file) — the entire mitigation layer becomes unnecessary.

Happy to share the before/after wrapper diff if useful. LGTM from a consumer POV; would love to see this merged.

Issue #976, opened 2026-05-10
No deduplication when requestId is empty (AntCC data)


xavier-niu:
## Bug

ccusage only deduplicates when both `messageId` AND `requestId` are present:

```ts
// data-loader.ts:530
export function createUniqueHash(data: UsageData): string | null {
  const messageId = data.message.id;
  const requestId = data.requestId;
  if (messageId == null || requestId == null) {
    return null;
  }
  return `${messageId}:${requestId}`;
}
```

When `requestId` is empty/null, all records skip dedup entirely.

## Impact

AntCC (CodeFuse engine) JSONL files always have empty `requestId`. This causes ccusage to overcount tokens by ~2.3x because the same message is written multiple times (streaming updates) with the same `messageId` but no `requestId`.

Example from real data:
- Raw records: 368
- Unique `messageId`: 171
- Duplicate entries: 197
- Input tokens (raw, no dedup): 1,071,492
- Input tokens (deduped by `messageId`): 465,781

## Suggested fix

When `requestId` is empty but `messageId` is present, deduplicate by `messageId` alone. The hash could be just `messageId` in that case, or `messageId:` (with empty requestId suffix).

```ts
export function createUniqueHash(data: UsageData): string | null {
  const messageId = data.message.id;
  const requestId = data.requestId;
  if (messageId == null) {
    return null;
  }
  return requestId != null ? `${messageId}:${requestId}` : messageId;
}
```

This preserves the existing behavior for Claude Code data (which always has both fields) while fixing the overcount for AntCC data.

xiufengsun:
Re-opening visibility on this — TokenTracker just shipped a fix for the same short-circuit (mm7894215/TokenTracker#65). Same root cause as #797.

We confirmed the field is **absent** (not `null`) for entries from DeepSeek `/anthropic`, Kimi anthropic mode, Mimo, MiniMax, and Claude Code sub-agent / `*-thinking` transport paths.

Same fix in TS:
```ts
export function createUniqueHash(data: UsageData): string | null {
  const messageId = data.message.id;
  if (messageId == null) return null;
  const requestId = data.requestId;
  return requestId != null ? `${messageId}:${requestId}` : messageId;
}
```

Validation: our author's deepseek-v4-flash today drops from 5,649,330 (raw sum) to 1,922,817 after dedup, vs DeepSeek's own console showing 2,250,037 (85.5% match, residual from TZ/buffer/timing).

[...]

Issue #888, opened 2026-03-11
Claude usage can be undercounted when the same request is logged multiple times


CloveSVG:
## Summary

`ccusage` can undercount Claude transcript usage when the same assistant response is written multiple times with the same `message.id` and `requestId`.

The current dedupe logic appears to keep the first seen entry for a given `message.id:requestId` pair and drop later entries. In recent Claude Code transcripts, the first entry is often an intermediate usage snapshot, while the last entry contains the final `output_tokens` for that same request.

This makes daily/session totals systematically too low, especially for `output_tokens` and calculated cost.

## Environment

- `ccusage` version: `18.0.10`
- command used: `npx ccusage daily --since 20260310 --until 20260310 --json --mode calculate`
- Claude transcript source: local `~/.claude/projects/**/*.jsonl`

## Minimal reproduction pattern

I am seeing transcript entries like this (redacted / minimized):

```json
{"type":"assistant","timestamp":"2026-03-11T02:40:21.204Z","requestId":"req_x","message":{"id":"msg_x","model":"claude-opus-4-6","usage":{"input_tokens":3,"cache_creation_input_tokens":2285,"cache_read_input_tokens":22137,"output_tokens":2}}}
{"type":"assistant","timestamp":"2026-03-11T02:40:22.650Z","requestId":"req_x","message":{"id":"msg_x","model":"claude-opus-4-6","usage":{"input_tokens":3,"cache_creation_input_tokens":2285,"cache_read_input_tokens":22137,"output_tokens":1093}}}
```

Both lines have the same `message.id` and `requestId`, but the later line is clearly the final usage snapshot.

If the first line is kept and the second line is discarded as a duplicate, the request is undercounted by `1091` output tokens.

## Why I think this is happening

The installed code path in `dist/data-loader-B58Zt4YE.js` builds a dedupe key from:

- `message.id`
- `requestId`

and then skips any later entry with the same key.

In practice, this means "first seen wins".

For these Claude transcripts, a safer rule seems to be one of:

1. keep the latest entry for the same logical request, or
2. keep the entry with the largest total usage, or
3. at minimum, prefer a later entry when usage differs.

## What I measured locally

I wrote a small one-off parser over the raw transcript JSONL files for a single day and compared three dedupe strategies for assistant entries grouped by `sessionId + requestId`:

- first seen: `outputTokens = 130,785`
- latest seen: `outputTokens = 648,562`
- max seen: `outputTokens = 648,649`

[...]

PR #985, opened 2026-05-12
fix(data-loader): dedup by messageId alone when requestId is absent

### LINKED ISSUE #797: Usage is wrong for duplicate rows per message
When there are duplicate rows per message, the tool is taking usage from the first row, while it should take it from the last row

xiufengsun:
Fixes #797.
Addresses #976.
Refs #705, #888.

## Summary

- Uses message.id as the fallback dedupe key when a Claude usage entry has no requestId.
- Keeps the existing messageId:requestId key when requestId is present.
- Adds regression coverage for the hash fallback and daily aggregation of duplicated requestId-less entries.

## Scope

This fixes the requestId-absent dedupe short-circuit only. It does not fix the separate first-wins streaming snapshot issue discussed in #705 and #888, where duplicated entries with the same key may need the final or largest usage record.

## Validation

- Real local Claude data scan: 3,144 JSONL files / 155,417 usage rows; 263 requestId-absent rows; no local token delta because the duplicated requestId-absent rows in this dataset have zero token totals.
- pnpm vitest run apps/ccusage/src/adapter/claude/data-loader.ts -t "request id is missing|same message id when request id is missing"
- pnpm run format
- pnpm typecheck
- pnpm run test

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deduplication for Claude usage records: entries that share a message identifier but lack a request identifier are now treated as duplicates, reducing duplicate daily usage entries.

* **Tests**
  * Expanded test coverage to validate the new deduplication fallback and daily-usage deduplication behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ryoppippi/ccusage/pull/985?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/ccusage/src/adapter/claude/data-loader.ts
+++ b/apps/ccusage/src/adapter/claude/data-loader.ts
@@ -1173,18 +1173,16 @@ async function filterFilesByMtime<T>(
 }
 
 /**
- * Create a unique identifier for deduplication using message ID and request ID
+ * Create a unique identifier for deduplication using message ID and request ID.
  */
 export function createUniqueHash(data: UsageData): string | null {
 	const messageId = data.message.id;
-	const requestId = data.requestId;
-
-	if (messageId == null || requestId == null) {
+	if (messageId == null) {
 		return null;
 	}
 
-	// Create a hash using simple concatenation
-	return `${messageId}:${requestId}`;
+	const requestId = data.requestId;
+	return requestId == null ? messageId : `${messageId}:${requestId}`;
 }
 
 async function processJSONLUsageFileByLine(
@@ -6204,7 +6202,7 @@ if (import.meta.vitest != null) {
 				expect(hash).toBeNull();
 			});
 
-			it('should return null when request id is missing', () => {
+			it('should fall back to message id when request id is missing', () => {
 				const data = {
 					timestamp: createISOTimestamp('2025-01-10T10:00:00Z'),
 					message: {
@@ -6217,7 +6215,7 @@ if (import.meta.vitest != null) {
 				};
 
 				const hash = createUniqueHash(data);
-				expect(hash).toBeNull();
+				expect(hash).toBe('msg_123');
 			});
 		});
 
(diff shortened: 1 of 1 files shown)


xiufengsun:
## Follow-up: independent user confirmation

Posting a real-world validation in case it helps move the review along.

The same fix shipped in TokenTracker v0.12.1 (mm7894215/TokenTracker#65) was just confirmed by an end user on mm7894215/TokenTracker#64 — they were the original reporter whose DeepSeek console screenshots surfaced the bug:

> 更新到最新版本后, 确实tokentracker和deepseek官方的计数一致了，不仅是今天，昨天的错误计数也修复了
>
> (After updating to the latest version, TokenTracker's count now matches DeepSeek's official console — not just today's, the previously-inflated historical counts are fixed too.)

So in addition to my own machine's before/after table in the PR description, there's now an independent user reproducing the fix against a third-party authoritative source (DeepSeek's billing console) on their own data.

Happy to add anything else that would help review — more test cases, additional provider coverage, a changelog entry, etc. Just let me know.

ryoppippi:
Maintainer check after rebasing this onto current main:

- Moved the fix to the current Claude adapter path after #1004: apps/ccusage/src/adapter/claude/data-loader.ts.
- Confirmed the regression test is red on main and green after the change.
- Added an aggregation regression test for duplicated entries with the same message.id and no requestId.
- Scanned local Claude data: 3,144 JSONL files / 155,417 usage rows; 263 rows had no requestId, but this local dataset had no non-zero token delta from the fix. So local data does not reproduce the overcount; the merge case rests on the external reports in #797/#976 and the low-risk fallback behavior.
- Updated the PR body so this no longer claims to fully fix #705/#888. Those are still the separate first-wins/final-usage issue.

Validation run locally:

- pnpm vitest run apps/ccusage/src/adapter/claude/data-loader.ts -t "request id is missing|same message id when request id is missing"
- pnpm run format
- pnpm typecheck
- pnpm run test

Decision: OK to merge as the narrow requestId-absent dedupe fix once bot checks finish; do not treat it as closing #705/#888.

xiufengsun:
Re your local scan finding zero token delta — that's expected, and points to *why* the bug shows up so unevenly across users. Sharing some data from my end in case it helps.

**TL;DR: This affects users who route Claude Code through third-party / OpenAI-compatible API gateways (DeepSeek, Kimi, Mimo, MiniMax, etc.). Anthropic-subscription users almost never see it.**

The reason: those gateways emit JSONL entries **without a top-level `requestId`** (it's an Anthropic-API-specific field). Their `message.id` is the upstream provider's id (e.g. `chatcmpl-...` from an OpenAI-compatible response). When Claude Code re-writes the same assistant message to the JSONL (which it does — I see 2-4× duplication routinely), `createUniqueHash` short-circuits to `null` and **every duplicate is counted**.

Scan from one machine (~5 months of mixed-provider Claude Code usage):

| Model | Pre-fix input tokens | Post-fix input tokens | Overcount |
|---|---:|---:|---:|
| kimi-for-coding | 41,489,695 | 23,971,493 | **+42.2%** |
| kimi-k2.5 | 1,837,702 | 583,130 | **+68.3%** |
| mimo-v2.5-pro | 744,629 | 165,838 | **+77.7%** |
| deepseek-v4-flash | 490,064 | 201,727 | **+58.8%** |
| MiniMax-M2.1 | 317,849 | 95,508 | **+70.0%** |
| **Total** | **44,879,939** | **25,017,696** | **+44.3% (+19.8M)** |

440 distinct `(model, message.id)` groups affected, each with 2-4 duplicates and identical token counts. Sanitized real entry (one of three identical writes for the same `message.id`):

\`\`\`json
{
  "type": "assistant",
  "timestamp": "2026-01-27T08:50:16.074Z",
  "message": {
    "id": "chatcmpl-69787c4119bd8881d3d733dc",
    "model": "kimi-k2.5",
    "role": "assistant",
    "usage": { "input_tokens": 25210, "output_tokens": 0 }
  }
}
\`\`\`

No \`requestId\` at the top level — that's the field ccusage was keying on.

An Anthropic-subscription user's JSONL has \`requestId\` populated on every assistant message, so the short-circuit path is almost never hit, and the 263 requestId-absent rows you saw locally are likely edge cases (synthetic responses, error replies, etc.) that genuinely had zero tokens. That matches your scan exactly.

Net: the fallback isn't just "low-risk" — it's the only thing keeping third-party-routed users' totals honest. Reproduction fixture in the new aggregation test is structurally identical to what these gateways emit.

ryoppippi (approved):
LGTM

Reviewed and validated the rebased change.

Head SHA: 7f01804ef1ef9a8fb55f1d2be7f2c13bc4ea2a25

This is the narrow Claude requestId-absent dedupe fix only. It keeps the existing messageId:requestId behavior when requestId is present and uses message.id as the fallback key when requestId is absent.

Local validation after rebase:

- pnpm vitest run apps/ccusage/src/adapter/claude/data-loader.ts -t "request id is missing|same message id when request id is missing"

PR #710, opened 2025-10-29
feat(ccusage): Add --prompts flag to display prompt counts across all time-based reports

antonioribeiro:
I know it sounds like a duplicate of https://github.com/ryoppippi/ccusage/pull/679, but this one is not for blocks ;) 

### Summary
Added a new `--prompts` flag to daily, weekly, and monthly reports that displays the number of individual prompts sent to Claude during each time period, helping users understand their usage patterns beyond just token counts and costs.

### Background
Previously, users could only see token counts and costs in usage reports. This feature adds prompt count tracking to provide additional insight into usage frequency and patterns.

### Changes Made

**Data Infrastructure (`data-loader.ts`)**:
- Added `promptCount: v.number()` to `dailyUsageSchema`, `weeklyUsageSchema`, and `monthlyUsageSchema`
- Updated `bucketUsageSchema` to include `promptCount`
- Modified `loadDailyUsageData()` to calculate `promptCount: entries.length`
- Enhanced `loadBucketUsageData()` to aggregate prompt counts from daily data
- Added comprehensive test coverage for prompt count aggregation across all time periods

**Terminal Package (`packages/terminal/src/table.ts`)**:
- Updated `UsageReportConfig` interface to include `includePrompts?: boolean`
- Enhanced `createUsageReportTable()` to support dynamic prompt column building
- Modified `formatUsageDataRow()` and `formatTotalsRow()` to handle prompt counts
- Added extensive test coverage for table formatting with prompts column
- Implemented responsive table layout that adapts to prompt column inclusion

**Daily Command (`commands/daily.ts`)**:
- Added `prompts` argument to command configuration (boolean flag, default: false)
- Updated table configuration to include `includePrompts: ctx.values.prompts`
- Modified data formatting to pass prompt counts to table rows
- Updated JSON output to include `promptCount` field
- Added prompt count calculation to totals row

**Weekly Command (`commands/weekly.ts`)**:
- Added `prompts` argument to command configuration (boolean flag, default: false)
- Updated table configuration to include `includePrompts: ctx.values.prompts`
- Modified data formatting to pass prompt counts to table rows
- Updated JSON output to include `promptCount` field
- Added prompt count calculation to totals row

**Monthly Command (`commands/monthly.ts`)**:
- Added `prompts` argument to command configuration (boolean flag, default: false)
- Updated table configuration to include `includePrompts: ctx.values.prompts`
- Modified data formatting to pass prompt counts to table rows
- Updated JSON output to include `promptCount` field
[...]

PR #679, opened 2025-09-30
feat(blocks): add --prompts flag to show prompt count per block

processtrader:
Add a new --prompts/-p flag to the blocks command that displays the number of prompts sent in each 5-hour billing block. This is particularly useful for GLM users and other models where prompt-based tracking is more relevant than token-based billing.

Changes:
- Add prompts flag to CLI arguments with short option -p
- Add "Prompts" column to table output when flag is enabled
- Show prompt count using block.entries.length
- Update gap rows, REMAINING rows, and PROJECTED rows for column alignment
- Update README with comprehensive documentation for GLM users
- Regenerate config schema to include new flag

## Summary by CodeRabbit

* **New Features**
  * Add --prompts (-p) flag to the blocks command to show a Prompts column with per-block prompt counts across all sections.
  * Add sessionLength and prompts options to the ccusage config to control per-session duration and enable prompt counting.
* **Documentation**
  * Update README with usage examples, rationale, GLM-specific notes, and sample outputs showing prompt counts and related metrics.

image

PR #992, opened 2026-05-12
feat(ccusage): add --prompts flag to claude daily/weekly/monthly reports (rebase of #710 onto v19)

akiratee:
## Summary

Reimplements @antonioribeiro's #710 (`--prompts` flag for daily/weekly/monthly reports) on top of the v19 adapter architecture. The original PR sat against pre-v19 code that has since been substantially reshaped (#1004 adapter unification, #1007 file-prefix rename, #1000 all-agent reports). A literal rebase wasn't viable, so this is a fresh implementation that preserves #710's scope and behavior while fitting the new layout.

Scope is intentionally limited to the **Claude adapter** to match #710. Generalizing `promptCount` to the shared adapter layer (so Codex/Amp/OpenCode/Pi inherit the column) is a separate change - happy to do that if you'd prefer it bundled.

## What's in here

**Adapter (`apps/ccusage/src/adapter/claude/data-loader.ts`)**
- Track `promptCount` on the shared `UsageSummaryAccumulator` and surface it from `finalizeUsageSummary`.
- Add `promptCount: number` to `DailyUsage`, `WeeklyUsage`, `MonthlyUsage`, and `BucketUsage`.
- Bucket aggregation (used by weekly/monthly) sums each daily's `promptCount` so the numbers stay consistent across reports.

**Terminal (`packages/terminal/src/table.ts`)**
- New `includePrompts?: boolean` on `UsageReportConfig`.
- New optional `promptCount?: number` on `UsageData`.
- When set, the `Prompts` column slots between Models (or Agent) and Input in both base and compact layouts.
- `formatUsageDataRow` and `formatTotalsRow` derive whether to render a Prompts cell from `data.promptCount != null` - no signature breakage for existing callers.

**Commands (`apps/ccusage/src/commands/{daily,weekly,monthly}.ts`)**
- New `prompts` boolean CLI arg (defaults to `false`).
- Thread `includePrompts` through table config, row/total formatters, separator column count, and `pushBreakdownRows` offset (2 instead of 1 when `--prompts` is on) so `--breakdown` rows stay aligned.
- JSON output gates `promptCount` on rows, totals, and `emptyOutput` totals behind `--prompts` so the payload shape is stable when the flag is off.

**Project grouping (`apps/ccusage/src/daily-grouping.ts`, `json-output-types.ts`)**
- `groupByProject` takes a new `includePrompts` argument and propagates `promptCount` into `DailyProjectOutput` (which gained an optional `promptCount` field). Keeps `daily --instances --prompts --json` consistent with the flat output.

**Misc**
- README: added a feature-list bullet describing the flag and its semantics.
- `config-schema.json` regenerated.

## Note on semantics

`promptCount` counts usage-bearing entries (one per assistant API call). A single user message that triggers N tool iterations counts as N+ entries, not 1. The companion PR #679 (blocks command) filters to user-role messages instead, so the two PRs count different things under the same flag. Worth deciding whether to unify the definition before merging both.

## Test plan

- [x] `pnpm test` - 498/498 pass (3 pre-existing skipped)
- [x] `pnpm lint` - clean
- [x] `pnpm --filter ccusage typecheck` - clean
- [x] `pnpm --filter ccusage build` - clean
- [x] Manually verified `ccusage claude daily` / `claude weekly` / `claude monthly` against real Claude Code data: Total-row input-token / cost figures match between flag states (proves no column shift).
- [x] Manually verified `ccusage claude weekly --breakdown --prompts`: breakdown rows have an empty Prompts cell so model token columns stay aligned with the data row above.
- [x] Verified `--json` gating: `promptCount` absent on rows and totals without `--prompts`, present with it.
- [x] Verified `ccusage claude daily --instances --prompts --json` propagates `promptCount` through `groupByProject` and totals.
- [x] Confirmed default `ccusage daily` (all-agent unified report) is unaffected.

Closes #710. Companion PR #679 is independent.

## Summary by CodeRabbit

* **New Features**
  * Added a --prompts option for daily, weekly, and monthly reports to show a Prompts column (counts of usage-bearing API requests, including tool-call iterations). Feature is disabled by default.
* **Documentation**
  * README updated with a Prompt Count feature note explaining the Prompts column and semantics.
* **Tests**
  * Updated tests/fixtures to cover prompt-count behavior in aggregation and output.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ryoppippi/ccusage/pull/992?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/ccusage/README.md
+++ b/apps/ccusage/README.md
@@ -94,6 +94,7 @@ bunx ccusage monthly --compact  # Compact monthly report
 - 🚀 **Statusline Integration**: Compact usage display for Claude Code status bar hooks (Beta)
 - 🤖 **Model Tracking**: See which models are used across supported sources
 - 📊 **Model Breakdown**: View per-model cost breakdown with `--breakdown` flag
+- ✉️ **Prompt Count**: Add a Prompts column to Claude daily/weekly/monthly reports with `--prompts` (counts usage-bearing entries - one per API request, including each tool-call iteration)
 - 📅 **Date Filtering**: Filter reports by date range using `--since` and `--until`
 - 📁 **Custom Paths**: Support for custom local data directory locations
 - 🎨 **Beautiful Output**: Colorful table-formatted display with automatic responsive layout
--- a/apps/ccusage/config-schema.json
+++ b/apps/ccusage/config-schema.json
@@ -165,6 +165,12 @@
 									"description": "Comma-separated project aliases (e.g., 'ccusage=Usage Tracker,myproject=My Project')",
 									"markdownDescription": "Comma-separated project aliases (e.g., 'ccusage=Usage Tracker,myproject=My Project')"
 								},
+								"prompts": {
+									"type": "boolean",
+									"description": "Show prompt count column (number of usage-bearing entries per bucket)",
+									"markdownDescription": "Show prompt count column (number of usage-bearing entries per bucket)",
+									"default": false
+								},
 								"all": {
 									"type": "boolean",
 									"description": "Accepted for compatibility; all detected supported agents are included by default",
@@ -256,6 +262,12 @@
 									"type": "boolean",
 									"description": "Force compact table layout for narrow terminals",
 									"markdownDescription": "Force compact table layout for narrow terminals",
+									"default": false
+								},
+								"prompts": {
+									"type": "boolean",
+									"description": "Show prompt count column (number of usage-bearing entries per bucket)",
+									"markdownDescription": "Show prompt count column (number of usage-bearing entries per bucket)",
 									"default": false
 								},
 								"all": {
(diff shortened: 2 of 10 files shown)


akiratee:
Heads up - main moved substantially since I opened this (v19 major cut, the adapter unification in #1004, file-prefix rename in #1007, the all-agent reports in #1000). The rebase isn't mechanical anymore - the data loader now lives in \`adapter/claude/data-loader.ts\` and command shapes changed enough that I'm going to re-implement on top of v19 rather than force-fit the old diff.

Keeping scope the same as #710 - \`--prompts\` for daily/weekly/monthly on the Claude adapter. Happy to generalize \`promptCount\` to the shared adapter layer so Codex/Amp/OpenCode/Pi get it for free, but that feels like a separate PR unless you'd prefer it bundled. Let me know.

Will push an updated branch shortly.

ryoppippi:

PR #1107, opened 2026-05-20
fix(ccusage): normalize dashed dates in --since/--until filters

Sim-hu:
## Summary

`--since` / `--until` silently produce wrong results when given a dashed `YYYY-MM-DD` date, even though the docs advertise that format alongside `YYYYMMDD`.

Date filtering compares entry dates with their dashes stripped (`20260315`) against the raw flag value. A dashed input like `2026-03-15` is compared verbatim, and since `-` (0x2D) sorts before the digits next to it, `--since` matches every row and `--until` matches none — with no error reported.

Reproduced with the release binary against real data (range 2026-01-23 – 2026-05-20):

| command | rows returned |
|---|---|
| `--since 20260315` | 29 |
| `--since 2026-03-15` | 39 (all rows — filter is a silent no-op) |
| `--until 20260315` | 10 |
| `--until 2026-03-15` | 0 (empty report) |

## Changes

Normalize `--since` / `--until` to the dashless form at parse time — for both CLI flags and config-file values — via a small `normalize_date_filter` helper. After the fix the dashed and dashless forms return identical results.

Adds unit tests for the helper and for dashed CLI-flag parsing, and updates one config test that asserted the previously-unnormalized value.

---
## Summary by cubic
Fixes date filters so dashed `YYYY-MM-DD` values in `--since`/`--until` are normalized and behave the same as `YYYYMMDD`. Prevents the prior bug where dashed input made `--since` match all rows and `--until` match none.

- **Bug Fixes**
  - Normalize `--since`/`--until` to `YYYYMMDD` at parse time for both CLI flags and config values.
  - Added `normalize_date_filter` helper with tests; updated config test expectations.
  - Both date formats now produce identical, correct results.

<sup>Written for commit 7ba845dfbeb52c6adfb3e9fd061855447e8a7a72. Summary will update on new commits. <a href="https://cubic.dev/pr/ryoppippi/ccusage/pull/1107?utm_source=github">Review in cubic</a></sup>

## Release Notes

* **Bug Fixes**
  * Date filter arguments (`--since` and `--until`) now properly normalize dates provided in dashed format (YYYY-MM-DD) for consistent date filtering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ryoppippi/ccusage/pull/1107?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/rust/crates/ccusage/src/cli.rs
+++ b/rust/crates/ccusage/src/cli.rs
@@ -712,8 +712,16 @@ fn parse_shared_arg_for_command(
 
 fn parse_shared_arg(parser: &mut ArgParser, shared: &mut SharedArgs) -> Result<(), String> {
     match parser.next_flag()?.as_str() {
-        "-s" | "--since" => shared.since = Some(parser.value_for("--since")?),
-        "-u" | "--until" => shared.until = Some(parser.value_for("--until")?),
+        "-s" | "--since" => {
+            shared.since = Some(crate::date_utils::normalize_date_filter(
+                &parser.value_for("--since")?,
+            ))
+        }
+        "-u" | "--until" => {
+            shared.until = Some(crate::date_utils::normalize_date_filter(
+                &parser.value_for("--until")?,
+            ))
+        }
         "-j" | "--json" => shared.json = true,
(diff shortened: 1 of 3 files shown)


ryoppippi:

Issue #867, opened 2026-02-25
feat(opencode): add --since / --until date filters for daily report


daheli:
### Problem

`@ccusage/opencode daily` currently does not support date range filtering.

`ccusage daily` already supports:

- `--since <YYYYMMDD>`
- `--until <YYYYMMDD>`

But `@ccusage/opencode daily --help` only shows `--json` and `--compact`, so users must pipe JSON to `jq` for date filtering.

### Why this matters

- Inconsistent UX between `ccusage` and `@ccusage/opencode`
- Harder scripting for common tasks like "last 14 days excluding today/yesterday"
- Extra dependency on `jq` for a very common use case

### Proposal

Add these options to `@ccusage/opencode daily`:

- `-s, --since <since>` (YYYYMMDD)
- `-u, --until <until>` (YYYYMMDD)

Optional (nice to have):
- support ISO date input (`YYYY-MM-DD`) as well, with clear validation errors

### Expected behavior

```bash
npx @ccusage/opencode@latest daily --since 20260211 --until 20260223
```

Returns only rows within the inclusive range.

### Notes

Current workaround:

```bash
npx @ccusage/opencode@latest daily --json | jq '...'\n```\n\nbut built-in filtering would be much more ergonomic.

ryoppippi:
Closing during repository triage: this is covered by current main or later merged work.

Issue #801, opened 2026-01-12
`--since` option does not filter dates for `@ccusage/opencode`


n2k3:
The `--since` option appears to have no effect when used with `@ccusage/opencode`. The report includes all historical data regardless of the specified date.

**Steps to reproduce**
All of the following commands produce identical, unfiltered output:
```bash
bunx @ccusage/opencode@latest daily --since 20260107
bunx @ccusage/opencode@latest daily --since 20260107 --compact
bunx @ccusage/opencode@latest daily --since 2026-01-07
bunx @ccusage/opencode@latest daily --since 2026-01-07 --compact
bunx @ccusage/opencode@latest --since 20260107
bunx @ccusage/opencode@latest --since 20260107 --compact
bunx @ccusage/opencode@latest --since 2026-01-07
bunx @ccusage/opencode@latest --since 2026-01-07 --compact
```

**Expected behavior**

The daily report should only show entries from 2026-01-07 onward (inclusive).

**Actual behavior**

<details>

<summary> 👉 click to show command log</summary>

```
paul@Mac virtoura % bunx @ccusage/opencode@latest daily --since 20260107 --compact

 WARN  Fetching latest model pricing from LiteLLM...

ℹ Loaded pricing for 2418 models

📊 OpenCode Token Usage Report - Daily

┌────────────┬─────────────────────────────────────────────────┬───────────┬────────────┬─────────────┐
│ Date       │ Models                                          │     Input │     Output │  Cost (USD) │
├────────────┼─────────────────────────────────────────────────┼───────────┼────────────┼─────────────┤
│ 2025-12-26 │ - big-pickle                                    │    96,797 │     46,984 │       $3.65 │
│            │ - glm-4.7-free                                  │           │            │             │
│            │ - sonnet-4-5                                    │           │            │             │
├────────────┼─────────────────────────────────────────────────┼───────────┼────────────┼─────────────┤
│ 2025-12-27 │ - sonnet-4-5                                    │        69 │     17,549 │       $1.10 │
├────────────┼─────────────────────────────────────────────────┼───────────┼────────────┼─────────────┤
│ 2025-12-28 │ - opus-4-5                                      │       251 │     34,367 │       $4.49 │
│            │ - sonnet-4-5                                    │           │            │             │
├────────────┼─────────────────────────────────────────────────┼───────────┼────────────┼─────────────┤
│ 2025-12-29 │ - opus-4-5                                      │        15 │     44,857 │      $14.24 │
[...]

PR #1188, opened 2026-05-29
fix(opencode): honor --since/--until in SQLite loader and JSON dump scan

justi:
## Summary

`ccusage opencode` (and the aggregate `ccusage`) reads every row from the OpenCode SQLite `message` table on every invocation, regardless of `--since` / `--until`, and then re-reads every JSON file under `storage/message/*.json`. On long-lived OpenCode installs (especially those migrated from the pre-SQLite layout) the loader can hang for minutes or appear to lock up.

`SharedArgs.since` / `SharedArgs.until` were already plumbed into the adapter (`ccusage-cli/src/types.rs:34-35`), but the loader did not use them.

## Reproduction (local data, before this PR)

On `main @ 9d90e1b`, opencode storage with:
- `opencode.db`: 33.6 GB, 392,340 rows in `message`
- `storage/message/`: 118,619 JSON files (legacy pre-SQLite dump, ~2.0 GB)

```bash
$ ccusage opencode --since 2026-05-04 --until 2026-05-10
# never returns; the loader scans 392k DB rows + parses 118k JSON files
```

Raw SQL with `WHERE time_created BETWEEN ? AND ?` against the same DB returns in <50 ms.

## Fix

`rust/crates/ccusage/src/adapter/opencode/loader.rs`:

1. **SQLite query uses the existing index.** The `message` table already has `CREATE INDEX message_session_time_created_id_idx ON message (session_id, time_created, id)`. When `shared.since` / `shared.until` are set, the prepared statement now adds `WHERE time_created >= ?1 AND time_created < ?2` (or open-ended variants).

2. **`storage/message/*.json` loop skips by mtime.** Files outside the inferred range are skipped via `fs::metadata(...).modified()` before any read/parse.

3. **Slack.** since/until are inflated by `-1` / `+2` days before being applied, so the existing string-based summary filter (`summary.rs:265-271`) remains authoritative; the SQL/mtime checks only short-circuit work that the summary would have discarded anyway. This absorbs timezone offsets and any FAT32-style 2-second mtime rounding.

## Performance on local data

| Variant | Runtime (`--since 2026-05-04 --until 2026-05-10`) |
|---|---:|
| `main @ 9d90e1b` | did not return |
| this PR, SQL `WHERE` only | 186.68 s |
| this PR, SQL `WHERE` + mtime-skip | 9.57 s |

Loaded tokens are identical across variants and match a hand-written `sqlite3` query against the same DB: 414,644 input / 4,646 output.

## Cross-OS compatibility

Runtime: `std::fs::metadata().modified()` works on macOS APFS/HFS+, Linux ext4/btrfs/xfs, and Windows NTFS. FAT32's 2-second mtime granularity is covered by the slack.

Tests: new tests use the `filetime` crate (`FileTime::from_unix_time`) which maps to `utimensat` (Linux), `setattrlist` (macOS), and `SetFileTime` (Windows). Same code path runs on all three CI targets.

## Tests added

4 new tests in `adapter::opencode::loader::tests`:
- `since_filter_drops_db_rows_older_than_lower_bound`
- `until_filter_drops_db_rows_at_or_after_upper_bound`
- `since_filter_skips_json_files_with_older_mtime`
- `no_since_until_keeps_all_json_files_regardless_of_mtime`

All 285 workspace tests pass; `cargo clippy --workspace --all-targets -- -D warnings` is clean; `cargo fmt --all -- --check` is clean.

## Why this is a fresh narrow PR

The same user-visible gap was reported in #801, requested in #867, and previously fixed against the now-removed TypeScript `@ccusage/opencode` package in #960 (closed with: *"If a gap remains, it needs a fresh narrow PR against main"*). This PR is that narrow PR against the current Rust adapter. No CLI surface, docs, or configuration schema changes.

---
## Summary by cubic
Make `ccusage opencode` honor `--since`/`--until` for both SQLite and JSON, avoiding full scans and cutting runtime on large installs from minutes to seconds.

- **Bug Fixes**
  - SQLite: push down indexed `WHERE time_created` bounds (open-ended variants) with bound params; fall back to an unfiltered scan if the column is missing, then apply a per-row date check.
  - JSON: extract `time.created` from raw content and skip out-of-range files before full parse; works with pretty-printed files and fails open to a full parse if extraction fails.
  - Safety: widen bounds by -1/+2 days so the summary-time filter stays authoritative and tolerates TZ drift.
  - Tests: add cases for SQL bounds, fallback path, and JSON content filtering.

<sup>Written for commit 24b7abe542b78bdd242ce2c2524aed9ce3f49819. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ccusage/ccusage/pull/1188?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added optional date-range filtering for data loading operations.
  * Implemented SQL-level query optimization for filtered database access.

* **Improvements**
  * Enhanced timestamp extraction from JSON data with fallback support.
  * Improved schema compatibility handling with automatic fallback scanning.

* **Tests**
  * Extended test coverage for date-range filtering across multiple scenarios.

### DIFF
--- a/rust/crates/ccusage/src/adapter/opencode/loader.rs
+++ b/rust/crates/ccusage/src/adapter/opencode/loader.rs
@@ -11,9 +11,9 @@ use super::{
     paths::paths,
 };
 use crate::{
-    LoadedEntry, PricingMap, Result,
+    LoadedEntry, PricingMap, Result, TimestampMs,
     cli::{CostMode, SharedArgs},
-    collect_files_with_extension, debug_log, parse_tz, read_files_parallel,
+    collect_files_with_extension, debug_log, format_date_tz, parse_tz, read_files_parallel,
 };
 
 pub(crate) fn load_entries(shared: &SharedArgs) -> Result<Vec<LoadedEntry>> {
@@ -151,12 +151,57 @@ fn load_entries_from_database(
         );
         return Vec::new();
     };
-    let Ok(mut statement) = connection.prepare("SELECT id, session_id, data FROM message") else {
-        debug_log(
-            shared,
-            format!("Failed to read OpenCode database: {}", db_path.display()),
-        );
-        return Vec::new();
+    let (since_ms, until_ms) = since_until_ms_with_slack(shared);
+    let sql = match (since_ms, until_ms) {
+        (Some(_), Some(_)) => {
+            "SELECT id, session_id, data FROM message \
+             WHERE time_created >= ?1 AND time_created < ?2"
+        }
+        (Some(_), None) => "SELECT id, session_id, data FROM message WHERE time_created >= ?1",
+        (None, Some(_)) => "SELECT id, session_id, data FROM message WHERE time_created < ?1",
+        (None, None) => "SELECT id, session_id, data FROM message",
+    };
+    let mut statement = match connection.prepare(sql) {
+        Ok(mut statement) => {
+            let bind_result = match (since_ms, until_ms) {
+                (Some(s), Some(u)) => statement.bind((1, s)).and_then(|()| statement.bind((2, u))),
+                (Some(s), None) => statement.bind((1, s)),
+                (None, Some(u)) => statement.bind((1, u)),
+                (None, None) => Ok(()),
+            };
+            if bind_result.is_err() {
+                debug_log(
+                    shared,
+                    format!(
+                        "Failed to bind date range to OpenCode query: {}",
+                        db_path.display()
+                    ),
+                );
+                return Vec::new();
+            }
+            statement
+        }
+        Err(_) => {
+            // Schema without `time_created` fails to prepare the filtered
+            // query. Fall back to an unfiltered scan instead of returning no
+            // rows; the in-loop date check below keeps results correct.
+            debug_log(
+                shared,
+                format!(
+                    "OpenCode database lacks a time_created column; scanning unfiltered: {}",
+                    db_path.display()
+                ),
+            );
+            let Ok(statement) = connection.prepare("SELECT id, session_id, data FROM message")
+            else {
+                debug_log(
+                    shared,
+                    format!("Failed to read OpenCode database: {}", db_path.display()),
+                );
+                return Vec::new();
+            };
+            statement
+        }
     };
     let mut entries = Vec::new();
     loop {
@@ -171,6 +216,11 @@ fn load_entries_from_database(
                 let Ok(data) = statement.read::<String, _>(2) else {
                     continue;
                 };
+                if let Some(millis) = extract_message_timestamp(&data)
+                    && !timestamp_within_range(millis, tz, shared)
+                {
+                    continue;
+                }
                 let Ok(value) = serde_json::from_str::<OpenCodeMessage>(&data) else {
                     continue;
                 };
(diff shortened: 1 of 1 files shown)


ryoppippi:
Thanks for the careful writeup — the motivation is solid and the SQL path is well reasoned. One correctness concern and a couple of robustness/nit items before this is safe to merge.

## Context

The authoritative date filter is `filter_loaded_entries_by_date` (`adapter/claude/mod.rs:121`), which runs *after* load and filters on each entry's **logical date** (`time.created` converted to TZ). So the new loader filtering is purely a load-time optimization, and the only correctness requirement is: **the loader must never drop an entry that `filter_loaded_entries_by_date` would keep.** Dropping extra rows just makes it slower; dropping kept rows silently under-counts.

## ✅ SQL `WHERE time_created` — looks correct

The `-1 day / +2 day` slack absorbs TZ offset (max ±14h < 1 day) and any small drift between the `time_created` column and JSON `data.time.created`, so it's always wider than the logical-date filter. No risk of dropping kept rows, and it rides the existing index. 👍

## ⚠️ JSON mtime skip on the `--until` side — silent under-count risk

File mtime is not guaranteed to match the message's logical time. `cp -r`, `rsync` without `-a`, backup restore, and cloud sync all reset mtime to "now". That's exactly the "legacy pre-SQLite dump migrated to a new machine" scenario this PR targets.

- `--since` (skip when `modified_ms < min`): if mtime is reset *newer* than real, the file is not skipped → no loss. Reasonably safe.
- `--until` (skip when `modified_ms >= max`): if a legacy file's mtime is reset to "now" but its logical date is in the past, an `--until <past date>` query will **skip the file even though its logical date is in range** → silent under-count.

"The summary filter is authoritative" does not protect this path, because the skip happens *before* load, so dropped rows never reach the authoritative filter. The new tests set mtime explicitly and so don't exercise the mtime-vs-logical-time divergence.

Suggested options:
- Drop the `--until` mtime skip (the `--since` skip alone should capture most of the 186s → 9.57s win; legacy data is old and rarely hits the upper bound anyway), or
- Make the mtime skip explicitly opt-in/heuristic and document that `--until` may under-count on environments where files were copied without preserving timestamps.

## ⚠️ DB schema without `time_created` — silent empty DB results

If an older/variant OpenCode schema lacks `time_created`, `connection.prepare(sql)` fails and the `else` branch returns `Vec::new()`. That means DB entries silently disappear **only when** `--since/--until` is set, while the no-filter path still works — an asymmetric regression. Consider falling back to the unfiltered query on prepare failure.

## Nits

- The JSON loop calls `fs::metadata(&file)` twice per file (once for since, once for until). With 118k files that's a lot of extra `stat`s — fold into a single metadata call.
- The two mtime blocks are near-duplicates; a small helper would read better.
- `filetime` as a dev-dependency only is fine.

Net: SQL path is good to go; the `--until` mtime skip and the prepare-failure fallback are the two things I'd want addressed first.

ryoppippi:
@justi heads up — #1220 (now closed) tackled the same issue with a different JSON-side strategy that I think is worth pulling into this PR.

Instead of skipping JSON files by file mtime, it extracts the real `time.created` millis straight from the raw JSON string and runs the *exact same* date check the authoritative `filter_loaded_entries_by_date` uses (`format_date_tz` → compare against since/until). That sidesteps the mtime concern I raised above entirely: it's exact, needs no `-1/+2 day` slack, and if extraction fails it falls through to a full parse (fails open), so it can never drop a row the authoritative filter would keep.

Your SQLite `WHERE time_created` push-down is the better half — it's the part that actually uses the index and gives the big speedup, which #1220 lacked. So the ideal is the combination:

- **SQLite**: keep your indexed `WHERE time_created` predicate (the real perf win).
- **JSON files**: replace the mtime skip with #1220's content-based `time.created` extraction + the shared date check.

That keeps the speedup while removing the silent under-count risk. Closing #1220 so this stays the single canonical PR — would you be up for folding that JSON approach in here? Happy to point you at the exact `extract_message_timestamp` / `timestamp_within_range` helpers from that PR.

justi:
Following up on your suggestion to fold #1220's JSON strategy into this PR.

I've combined both halves as you outlined:
- **SQLite**: kept the indexed `WHERE time_created` push-down with the -1d/+2d slack (the real index speedup).
- **JSON files**: replaced the mtime skip with #1220's content-based `time.created` extraction, running the same `format_date_tz` check as the authoritative `filter_loaded_entries_by_date` and failing open to a full parse. This removes the mtime under-count risk you raised.

I also addressed the prepare-failure fallback: a schema without `time_created` now falls back to an unfiltered scan (the in-loop date check still applies) instead of returning no rows.

Verified on a real 33 GB / 395k-row `opencode.db`: the `time_created` column equals the payload `time.created` in all 395,289 rows (max delta 0 ms), so the push-down can't drop a row the authoritative filter would keep. A one-month `--since/--until` window returns correct totals; the full unit suite, clippy, and fmt are clean.

Would you be willing to reopen this (or drop an `lgtm`) so I can push the combined version?

ryoppippi:
@justi plz! go ahead

justi:
@ryoppippi gentle nudge - the combined version (indexed SQL push-down + content-based JSON extraction) is pushed and ready, rebased on current main with the parallel file reads. The CI workflows still seem to be waiting on a maintainer "Approve and run" before they can start. Happy to address anything once they're green. Thanks!

ryoppippi:
Thanks for this. It landed as part of #1492 (6d53c575), and your commit is preserved there — the squash carries `Co-authored-by: Justyna Wojtczak`.

This PR identified the gap and the shape of the fix, and #1492 keeps that shape: push the window into SQL, skip the parse for out-of-range payloads, and leave the report's string comparison authoritative. What changed on top of it:

- **A panic on multi-byte bounds.** `ccusage opencode --since "abあcde"` aborted with `byte index 4 is not a char boundary`, and through the aggregate `ccusage` command it surfaced as `agent loader panicked`. The bounds are now parsed byte-wise, so a value that is not eight ASCII digits simply leaves that side of the window open.
- **The slack.** Rather than widening by a day to absorb timezone skew, the bounds now resolve to instants in the reporting timezone, so the SQL window and the report's window are the same window. A one-day margin does survive, but for a different reason: `time_created` is only a proxy for the payload timestamp the report actually filters on.
- **The index.** `WHERE time_created >= ?` cannot seek `(session_id, time_created, id)` — its leading column is `session_id`, and `EXPLAIN QUERY PLAN` reports `SCAN message`. Moving the bound into a subquery that selects `id` alone lets that index answer it as a covering scan. On a synthetic 200k-row / 0.83 GB fixture, the database part of a 7-day window went from ~1.1 s to ~34 ms.
- **A regression we both missed.** Because the loader now narrows while reading, an out-of-range query made an installed OpenCode vanish from the aggregate report's `Detected:` list. Detection asks the filesystem instead.
- **The raw-text scan.** It could latch onto a `"time"` occurring elsewhere in the payload. A failed scan is harmless — the full parse decides — but a wrong one silently drops an in-range entry, so the search is now confined to the message's own `time` object.

The mtime-based file skipping was dropped too: mtime is not the payload date once a dump has been copied or restored. Files are still read, but out-of-range ones are no longer parsed.

Closing this in favor of #1492. Thanks again for the report, the reproduction against a 33 GB install, and the benchmark numbers — they made the problem straightforward to confirm and fix.

ryoppippi:

PR #1228, opened 2026-06-06
ci(nix): cache the dev-shell closure and drop unused cachix

ryoppippi:
## Summary

`setup-nix` took ~200s per CI job and the Nix store cache was effectively useless. This makes the `cache-nix-action` cache actually work, removes the dead `cachix-action` steps, and bounds the cache budget.

## The problem

Profiling a recent run ([#27076169962](https://github.com/ryoppippi/ccusage/actions/runs/27076169962)), each `setup-nix` step spent ~200s, broken down as:

| sub-step | time |
|---|---|
| Install Nix | ~5s |
| Setup Cachix (numtide) | ~3s |
| **Cache restore (`cache-nix-action`)** | **~76s** |
| **`nix develop --command true`** | **~113s** |
| ↳ build `ccusage-deps` (crane Rust deps) | ~81s |
| ↳ build `generate-config-schema` + treefmt | ~11s |
| ↳ pnpm install (shellHook) | ~10s |

The cache *hit* (exact key, 2.1 GB restored) yet `ccusage-deps` rebuilt for 80s every run. Root cause: the `cache-nix-action` primary-key only hashed `flake.lock`, but the dev-shell closure also depends on the Rust toolchain and the crane `ccusage-deps` derivation (keyed on `Cargo.lock`, the Cargo manifests, `rust-toolchain.toml`, and the `version` from the root `package.json`). With none of those in the key:

1. `flake.lock` unchanged → key hit → stale closure restored
2. current `Cargo.lock` differs → `ccusage-deps` not present → 80s rebuild
3. exact-key hit → `cache-nix-action` skips the save → fresh deps never cached
4. repeat forever until `flake.lock` changes

Separately, the `cachix-action` steps were dead weight: the `ryoppippi` cache was only push-configured in the job that builds `ccusage-static` (musl), so the glibc dev-shell artefacts were never pushed, and the pull step was gated behind the push token so other jobs couldn't substitute from it either.

## What changed

- **Fix the cache key** to hash every input that changes the dev-shell closure: `flake.lock`, `flake.nix`, `default.nix`, `package.nix`, `nix/**/*.nix`, root `package.json` (version → crane drv name), `rust-toolchain.toml`, `rust/Cargo.lock`, `rust/**/Cargo.toml`. `pnpm-lock.yaml` and the workspace `package.json` files are deliberately excluded (no Nix derivation consumes JS deps — pnpm install runs in the shellHook into `node_modules`, outside the Nix store); this is documented inline.
- **Remove both `cachix-action` steps** and the unused `cachix-auth-token` input (plus its callers in `ci.yaml` and `release.yaml`), consolidating on `cache-nix-action`.
- **Bound the cache budget**: `gc-max-store-size-linux` caps each of the two Linux caches at 3.5 GB (the closure is a gc root, so it survives the trim), and per-run `purge` removes superseded keys so the more frequently rotating key can't push the repo past the 10 GB Actions cache limit. Nix caching only runs on Linux (arm64 + x64); macOS/Windows are untouched and keep their existing small Cargo `actions/cache`.
- **Add a weekly `Cache GC` workflow** that deletes caches not accessed in the last 14 days, reclaiming caches left behind by merged/abandoned branches.

## Why

After the first run on a new key rebuilds and *saves* the closure, subsequent runs restore it and skip the ~80s `ccusage-deps` rebuild (and the redundant ~3s cachix install + ~76s double-cache overhead is gone), which should bring `setup-nix` down substantially on warm runs.

## Testing

- `actionlint` + `zizmor` (high/high) clean on the new/changed workflow and action files
- `treefmt --fail-on-change` clean
- Real timing impact will be visible on this PR's own CI runs

---
## Summary by cubic
Speed up CI by making the Nix dev‑shell cache actually hit, removing unused Cachix, and adding a weekly cache GC. Warm runs now skip the ~80s `ccusage-deps` rebuild; purge and GC are tuned to avoid evicting fresh caches.

- **Refactors**
  - Fix `nix-community/cache-nix-action` key to hash all dev‑shell inputs (`flake.lock`, `flake.nix`, Nix files, Rust toolchain/Cargo manifests, root `package.json` version); JS deps excluded by design.
  - Bound cache size and purge superseded keys after 2 days to stay under the 10 GB Actions limit; caching runs on Linux only.
  - Remove `cachix/cachix-action` steps and the `cachix-auth-token` input; callers updated to use only `cache-nix-action`.

- **New Features**
  - Add a weekly Cache GC workflow that deletes caches not accessed in 14 days.
  - Sort caches by last accessed ascending so the sweep deletes the stalest entries first.

<sup>Written for commit 5a85b7906b4390a5313e0d6b2d78a477bb7d9a2c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ryoppippi/ccusage/pull/1228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Chores**
  * Improved build caching with broader hashing of Nix and Rust inputs for more reliable cache hits.
  * Added a weekly cache garbage-collection workflow to remove stale caches automatically.
  * Removed the external cache authentication step, simplifying CI configuration.
  * Adjusted build workflow to run the Nix setup step only for Linux builds.

### DIFF
--- a/.github/actions/setup-nix/action.yaml
+++ b/.github/actions/setup-nix/action.yaml
@@ -1,9 +1,5 @@
 name: Setup Nix
-description: Install Nix and configure Cachix
-inputs:
-  cachix-auth-token:
-    description: Cachix authentication token
-    required: false
+description: Install Nix and restore the Nix store from the GitHub Actions cache
 runs:
   using: composite
   steps:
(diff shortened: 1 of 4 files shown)


ryoppippi:
@coderabbitai please review this PR.

This reworks the Nix store caching in CI: it fixes the `cache-nix-action` primary-key so the dev-shell closure is actually cached (the previous key only hashed `flake.lock`, so `ccusage-deps` rebuilt ~80s every run and the stale cache was never re-saved), removes the dead `cachix-action` steps, bounds the cache budget with `gc-max-store-size` + per-run `purge`, and adds a weekly cache GC workflow.

I'd especially appreciate a careful look at:
- the `gc-max-store-size-linux` value (3.5 GB) and the 10 GB-budget reasoning
- whether the `purge` config could ever delete the just-saved primary-key cache
- the `cache-gc.yaml` permissions and the `gh cache delete` loop

ryoppippi:
@coderabbitai thanks for the thorough review — all three points addressed.

1. **`--limit 100` ordering bug** — fixed in `dca00f5`. Added `--sort last_accessed_at --order asc` so the sweep fetches the stalest caches first instead of the freshest. Verified the flags against `gh cache list --help` (default order is indeed `desc`).
2. **Within-7-day rotation accumulation** — addressed in `5a85b7`. Dropped `purge-created` from 7 days to 2 days so repeated `Cargo.lock` bumps in one week can't pile up multiple ~2.5 GB caches per arch. The active cache is always the current primary-key, which `purge-primary-key: never` protects regardless of age.
3. **3.5 GB cap / `purge-primary-key` / permissions** — no change needed per your ✅.

Please re-review the two follow-up commits when you get a chance.

PR #1287, opened 2026-06-11
ci: pin survival on build-native disk (do not merge)

ryoppippi:
Throwaway: after setup-nix (disk restored, before build) on the build-native job, check whether the pinned ccusage-deps profile and the musl deps path survived the sticky-disk trim. Determines if the profile pin actually works for the large build-native store or only worked for check by luck (small store, no trim).

---
<a href="https://app.blacksmith.sh/ccusage/codesmith/ccusage/pr/1287"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is enabled.</sup>

---
## Summary by cubic
Add a debug step to the `build-linux-native-package` action to verify the pinned `ccusage-deps` profile and musl deps survive the build-native disk restore/trim, without failing CI (no exit 1).
It lists `/nix/var/nix/profiles`, resolves the pin, checks `.#ccusage-static.cargoArtifacts.outPath` via `nix eval`/`nix-store`, and prints surviving GC roots.

<sup>Written for commit 17f937528f5204aee5437b953c170d75a27947ac. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ccusage/ccusage/pull/1287?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/actions/build-linux-native-package/action.yaml
+++ b/.github/actions/build-linux-native-package/action.yaml
@@ -36,7 +36,6 @@ runs:
         echo "::group::surviving gc roots"
         nix-store --gc --print-roots 2>/dev/null | grep -Ev '/proc/|/run/' | head -40 || true
         echo "::endgroup::"
-        exit 1
 
     - name: Build Linux static binary
       shell: bash
--- a/.github/actions/build-linux-native-package/action.yaml
+++ b/.github/actions/build-linux-native-package/action.yaml
@@ -15,6 +15,29 @@ runs:
   steps:
     - uses: ./.github/actions/setup-nix
 
+    - name: DEBUG did the pin survive on the build-native disk
+      shell: bash
+      run: |
+        echo "::group::profiles on restored disk"
+        ls -la /nix/var/nix/profiles/ 2>&1 || true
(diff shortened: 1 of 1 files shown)


ryoppippi:
Definitive result: the deps pin does NOT survive on the build-native sticky disk — the /nix/var/nix/profiles/ccusage-deps symlink persists but its target is GC'd; only the nix-quick-install user profile is honored by Blacksmith's trim. check warmed only because its store is small enough to skip the trim, not because the pin worked. cache-nix-action (#1285) works because its GC is standard 'nix store gc' which respects the profile root. Going with #1285.

PR #1414, opened 2026-07-09
feat(pricing): support OpenAI two-stage pricing and add the gpt-5.6 family

ryoppippi:
## Summary

OpenAI introduced two-stage (short/long context) pricing with the gpt-5.6 family: requests with more than 272K input tokens are billed at higher long-context rates. The same tier structure also applies to gpt-5.5, gpt-5.5-pro, gpt-5.4, and gpt-5.4-pro on the current pricing page. This PR adds the gpt-5.6-sol/terra/luna models to the built-in pricing table and makes Codex reports bill long-context requests at the correct rates.

## What Changed

- **Per-model tier threshold**: `Pricing` gains `long_context_threshold`; the previous hardcoded 200K boundary stays the default for LiteLLM `*_above_200k_tokens` data, while the OpenAI models switch tiers above 272K input tokens.
- **New built-in entries**: gpt-5.6-sol ($5/$30, long $10/$45), gpt-5.6-terra ($2.50/$15, long $5/$22.50), and gpt-5.6-luna ($1/$6, long $2/$9) per 1M input/output tokens, including cached-input and cache-write rates for both tiers.
- **Long-context overlay**: tier rates that upstream sources do not publish are re-applied after every pricing load via `builtin_long_context_rates`. A live LiteLLM refresh replaces whole entries (LiteLLM currently lists gpt-5.5/gpt-5.4 with flat rates only), so tier rates set directly on built-in entries would be silently dropped once a refresh succeeds. Upstream tier data wins over the overlay when it exists, and date-pinned keys such as `gpt-5.5-2026-04-23` share their base model's rates.
- **Codex per-request tiering**: OpenAI decides the tier per request and bills the whole request (input, cached input, and output) at long-context rates. Codex costs are computed from per-model sums, so `CodexModelUsage` now tracks the long-context portion while events are aggregated (each `token_count` event is one request), and `calculate_codex_model_cost` prices the short and long buckets independently. Models without tier rates are priced exactly as before.

## Notes

- Report JSON shape and table output are unchanged; only `costUSD` for long-context usage differs.
- The generic cost path (Claude Code, pi, and other non-Codex adapters) now mirrors the Codex per-request tiering: models with a per-model `long_context_threshold` bill the whole request at the long-context rates once input exceeds the threshold, so it is no longer a marginal breakpoint. LiteLLM `*_above_200k_tokens` data (no per-model threshold) keeps its marginal above-threshold semantics at the 200K default.
- Codex aggregation derives each request's tier from the matched model's threshold (`long_context_split_threshold`) instead of a single hardcoded 272K constant, so any model whose long-context boundary is not 272K is split at its own threshold.
- gpt-5.6 context limits mirror the 1,050,000-token window of gpt-5.5/gpt-5.4 until upstream metadata lands.
- `pricingOverrides` cannot override the tier threshold yet; that can be added later if someone needs it.

## Testing

- `just test` (Rust workspace + Node) passes: 343 + 61 + 16 + 11 crate tests, 0 failures.
- New tests: embedded gpt-5.6 rates and 272K threshold, overlay survival across a simulated LiteLLM refresh (and deferring to upstream tier data), date-suffix stripping, per-request long-context split during Codex aggregation (fixture-backed), two-stage Codex cost math, and flat-model equivalence.
- `cargo clippy -p ccusage --tests` and `just fmt` are clean.

---
<a href="https://app.blacksmith.sh/ccusage/codesmith/ccusage/pr/1414"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is enabled.</sup>

---
## Summary by cubic
Adds support for OpenAI two-stage pricing (short vs. long context) and the gpt-5.6 family. Codex now bills any request with more than 272K input tokens at long-context rates while keeping report shapes unchanged.

- **New Features**
  - Per-model tier threshold in pricing. Default stays 200K; OpenAI models use 272K.
  - Added gpt-5.6-sol, gpt-5.6-terra, and gpt-5.6-luna with embedded short-context and cache rates.
  - Filled long-context tier rates via a built-in overlay applied after every pricing load. Upstream tier data wins, and date-pinned keys inherit their base model’s rates.
  - Codex aggregation tracks long-context usage per request and prices short and long buckets separately. Models without tier rates are unchanged.
  - Set a 1,050,000-token context window for the gpt-5.6 family. Only costUSD changes for long-context usage; reports remain the same.

<sup>Written for commit 7d66785e5a0d0dafadbe15fd117590af0f0d1d62. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ccusage/ccusage/pull/1414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added long-context token tracking and pricing support, with separate accounting for long-context input/cached input/output.
  * Cost calculations now apply long-context/two-stage rates using model-specific thresholds.

* **Bug Fixes**
  * Improved aggregation to correctly preserve long-context usage when combining multiple requests.
  * Updated billing logic to better match tiered long-context behavior across supported model pricing sources.

* **Tests**
  * Expanded unit coverage for long-context split/threshold handling, tier switching, and aggregation correctness.

### DIFF
--- a/rust/crates/ccusage/src/adapter/codex/aggregate.rs
+++ b/rust/crates/ccusage/src/adapter/codex/aggregate.rs
@@ -336,8 +336,11 @@ fn accumulate_codex_event_into_group(
     model_usage.reasoning_output_tokens += event.reasoning_output_tokens;
     model_usage.total_tokens += event.total_tokens;
     // Each event is one request, so its input size decides the pricing tier
-    // here; the summed totals cannot recover per-request context sizes.
-    if event.input_tokens > crate::pricing::OPENAI_LONG_CONTEXT_THRESHOLD_TOKENS {
+    // here; the summed totals cannot recover per-request context sizes. The
+    // boundary is per model (OpenAI's 272K models and any future tier with a
+    // different threshold) rather than a single global constant, matching the
+    // threshold used to price the long-context buckets.
+    if event.input_tokens > crate::pricing::long_context_split_threshold(model) {
         model_usage.long_context_input_tokens += event.input_tokens;
         model_usage.long_context_cached_input_tokens += event.cached_input_tokens;
         model_usage.long_context_output_tokens += event.output_tokens;
--- a/rust/crates/ccusage/src/adapter/codex/aggregate.rs
+++ b/rust/crates/ccusage/src/adapter/codex/aggregate.rs
@@ -335,6 +335,13 @@ fn accumulate_codex_event_into_group(
     model_usage.output_tokens += event.output_tokens;
     model_usage.reasoning_output_tokens += event.reasoning_output_tokens;
     model_usage.total_tokens += event.total_tokens;
+    // Each event is one request, so its input size decides the pricing tier
+    // here; the summed totals cannot recover per-request context sizes.
+    if event.input_tokens > crate::pricing::OPENAI_LONG_CONTEXT_THRESHOLD_TOKENS {
+        model_usage.long_context_input_tokens += event.input_tokens;
+        model_usage.long_context_cached_input_tokens += event.cached_input_tokens;
+        model_usage.long_context_output_tokens += event.output_tokens;
+    }
     model_usage.is_fallback |= event.is_fallback_model;
 }
 
@@ -415,6 +422,9 @@ fn merge_groups(target: &mut BTreeMap<String, CodexGroup>, source: BTreeMap<Stri
             target_usage.output_tokens += usage.output_tokens;
             target_usage.reasoning_output_tokens += usage.reasoning_output_tokens;
             target_usage.total_tokens += usage.total_tokens;
+            target_usage.long_context_input_tokens += usage.long_context_input_tokens;
+            target_usage.long_context_cached_input_tokens += usage.long_context_cached_input_tokens;
+            target_usage.long_context_output_tokens += usage.long_context_output_tokens;
             target_usage.is_fallback |= usage.is_fallback;
         }
     }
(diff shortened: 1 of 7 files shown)


ryoppippi:
Pre-PR Codex review (`codex exec review --base main`, gpt-5.5) reported two P2 findings; both are deliberate decisions:

1. **Generic cost path keeps marginal above-threshold tiering.** Only the Codex adapter has verified per-request token semantics (`input_tokens` includes cached input), which the whole-request tier decision needs. Other adapters' token mappings were not audited in this PR, so their behavior is unchanged apart from the per-model threshold. Called out in the PR notes.
2. **The long-context overlay defers to upstream entirely once any tier rate exists.** Filling missing fields independently would mix upstream rates that assume the 200K LiteLLM boundary with built-in rates that assume the OpenAI 272K boundary under one per-model threshold, mispricing both tiers. Rationale recorded in the code comment (7d66785).

Issue #1436, opened 2026-07-12
Correct missing speed-tier rates and default pricing aliases


camjac251:
### What do you want to change?

Correct two cost-accounting paths in usage reports. When a speed tier has no published multiplier, retain standard pricing instead of assuming 2x. Also resolve a documented unsuffixed pricing family to its default entry before fuzzy matching, while preserving exact and partial user overrides.

The branch also marks golden snapshots as text and removes an unused timing-inference scratch plan.

### Why?

The generic multiplier can overstate cost without supporting rate data. Longest-name fuzzy matching can select a non-default sibling, and partially specified overrides can lose inherited fields. Both paths can misstate reported cost.

### How? (optional)

Use exact raw entry, canonical alias, then fuzzy fallback for pricing and context metadata. Canonicalize threshold lookups and inherit unspecified values from the canonical base entry. Add regression coverage for rates, context limits, thresholds, and override inheritance.

camjac251:
Update: #1437 is outdated. It was auto-closed before the source branch gained the inherited-history accounting fix, and GitHub froze that PR at commit `2500c5b`.

I pushed the complete four-commit branch at `b9cd694` and opened #1438 as the replacement. It contains the inherited child-session replay correction together with the speed-tier and pricing-alias fixes. Focused and full test suites, formatting, type checks, flake checks, and secret scanning pass; all committed fixtures use synthetic data.

#1438 was also auto-closed by the contributor gate. Please use #1438 for review and reopening; #1437 can remain closed.

camjac251:
@ryoppippi could this be merged?

ryoppippi:
hey! so does codex start recording fast/standard mode? they didn't record it before afaik.

@camjac251

camjac251:
Yes. Starting with CLI 0.144.0, rollouts persist `event_msg` → `thread_settings_applied` → `thread_settings.service_tier`. The recorded values are `priority` for Fast and `default` for Standard. I confirmed this against current upstream source and 0.144.x rollouts.

There is one limitation: usage before the first persisted settings snapshot can remain unclassified. #1438 now consumes explicit tier markers chronologically, keeps `--speed fast|standard` as full-report overrides, and uses the existing configuration fallback only for unclassified startup, legacy, and headless usage. It also prevents replayed parent settings from leaking into child usage and reconciles copied records deterministically before cost attribution.

[...]

Issue #1434, opened 2026-07-12
Significant accounting issues for codex


MilesCranmer:
### What happened?

I know this will get auto-closed but just wanted to let you know there are currently major accounting issues with codex for the past month @ryoppippi. I think it has something to do with subagents having all historical transcripts leading up to them get counted. For some days it seems like my usage is even 10x or 20x higher counting than expected. It might have something to do with recent codex updates, I am not too sure.

I'm looking into it now because it is deteriorating my personal tooling. If I succeed I can make a PR. Just wanted to make an issue to give a heads up.

There are other issues too for this. They were all auto-closed:

- #1349 (June 21, 2026)
- #1337 (June 18, 2026)
- #1288 (June 11, 2026)
- #950 (April 19, 2026)

### Steps to reproduce

N/A

### Version

20.0.17

ryoppippi:
thank you for reporting? could you investigate it?

MilesCranmer:
Will do! I've got a working patch (according to Codex's state_5.sql, it matches better) and am currently just trying to minify the diff.

MilesCranmer:
Hi @ryoppippi, here is my fix: #1435. (I think the auto-close is active unless you write `lgtmi` or something?)

yeqiugt:
this is very important to me, hope to merge quickly. Thank you.

axisrow:
+1 from real data. On my history (1,524 rollout files; **122 `forked_from_id`, 97 `thread_spawn`**), the giant "phantom" sessions on v20.0.17 are largely the parent's replayed `token_count` history being counted again inside the fork/subagent files — `#1369`'s same-second heuristic misses replays that span 2+ seconds, which is most of them.

@MilesCranmer's `#1435` is the right structural fix (parent→child matching via `session_meta` + prefix skip by delta equality). I reviewed it in [#1435](https://github.com/ccusage/ccusage/pull/1435) — it holds up, with one small test suggestion.

@ryoppippi, could you reopen `#1435`? It was auto-closed (new contributor), and this is actively deteriorating people's local tooling.

ryoppippi:
ok i'll take a look. thanks

ryoppippi:
lgtm

ryoppippi:
Historical audit: this discussion was auto-closed by the legacy contributor gate. That closure did not assess technical importance.

Audit result: **resolved**. A later merged change or the current `main` implementation covers this request. This item is kept for history and does not need to be reopened.

PR #1437, opened 2026-07-12
fix(usage): correct replay and pricing accounting

camjac251:
## Summary

Correct usage reports that can count inherited child-session history more than once, and fix several pricing paths involving speed tiers, default family aliases, and partial overrides.

Current child-session rollouts now establish an inherited baseline before counting the child's advancing usage. Unknown speed-tier rates retain standard pricing, while known aliases resolve to their documented default entries before fuzzy matching.

## What Changed

- exclude inherited parent-history prefixes from current child-session usage
- retain timestamp-based compatibility handling for older replay formats
- suppress repeated request deltas when cumulative totals do not advance
- preserve usage produced by the child around the durable turn boundary
- require an explicit multiplier before adjusting speed-tier costs
- resolve known family aliases before fuzzy pricing and context lookup
- preserve exact overrides and merge partial overrides onto canonical pricing
- mark golden snapshots as text and remove obsolete timing-inference notes
- add regression coverage for nested replay, legacy fallback, pricing, aliases, thresholds, and overrides

## Notes

- Report schemas and table layouts are unchanged.
- Unknown speed-tier rates remain at the standard estimate until an explicit multiplier is available.
- Exact pricing entries continue to take precedence over aliases and fuzzy matching.
- All committed fixtures use synthetic identifiers, timestamps, paths, and token values.

## Testing

- `nix --extra-experimental-features 'nix-command flakes' develop -c cargo test --manifest-path rust/Cargo.toml --workspace adapter::codex`
- `nix --extra-experimental-features 'nix-command flakes' develop -c just test`
- `nix --extra-experimental-features 'nix-command flakes' develop -c just typecheck`
- `nix --extra-experimental-features 'nix-command flakes' flake check`
- `git diff --cached --check`
- staged secret scan
- local-data smoke comparison without committing source records

Related: #1434, #1436


ryoppippi:

PR #1369, opened 2026-06-25
fix(codex): skip forked session replay history

sijie-ni-0214:
Summary:
- Treat Codex Desktop forked session logs as replay candidates when they contain `forked_from_id` metadata.
- Reuse the existing replay-prefix skipping path so copied parent `token_count` history is not counted again under the fork timestamp.
- Add a regression test covering forked sessions whose parent history timestamps are rewritten to the fork creation time.

Testing:
- `cargo fmt --manifest-path /Users/zilliz/.config/superpowers/worktrees/ccusage/fix-codex-fork-usage-dedupe/rust/Cargo.toml --all --check`
- `cargo test --manifest-path /Users/zilliz/.config/superpowers/worktrees/ccusage/fix-codex-fork-usage-dedupe/rust/Cargo.toml --workspace`
- `cargo run --quiet --manifest-path /Users/zilliz/.config/superpowers/worktrees/ccusage/fix-codex-fork-usage-dedupe/rust/Cargo.toml -p ccusage --bin ccusage -- codex daily --since 2026-06-25 --until 2026-06-25 --json --no-cost`

---
<a href="https://app.blacksmith.sh/ccusage/codesmith/ccusage/pr/1369"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784952251&installation_id=139163553&pr_number=1369&repository=ccusage%2Fccusage&return_to=https%3A%2F%2Fgithub.com%2Fccusage%2Fccusage%2Fpull%2F1369&signature=40b351209061d8a6aa6bf0d07fb83de4acb733f33b2dbb25fd4f66627b2cdc37"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
[...]

PR #1457, opened 2026-07-19
fix(codex): detect structural fork replay boundaries

MaxGhenis:
## Summary

- replace the one-second token heuristic with the Codex fork structure: adjacent leaf/ancestor `session_meta` records identify copied history, and `task_started.started_at` versus the record clock identifies the native boundary
- skip inherited prefixes even when they span multiple seconds, while carrying their terminal cumulative usage forward as the live baseline
- retain genuine fork-local events that share a second, and cover both modern `last_token_usage` and legacy cumulative-only rollouts
- preserve the same-second parent/child behavior established by [#1218](https://github.com/ccusage/ccusage/pull/1218) and extended to `forked_from_id` by [#1369](https://github.com/ccusage/ccusage/pull/1369)

The reproduction and expected arithmetic follow the independent fixture/oracle approach documented in the [Logpile accounting review](https://github.com/MaxGhenis/logpile/blob/main/docs/reviews/2026-07-11-sol-full-review.md). The exact 2026-07-18 matrix behind this change is published as the [2026-07-18 ccusage verification](https://github.com/MaxGhenis/logpile/blob/main/docs/reviews/2026-07-18-ccusage-verification.md).

## Testing

- `cargo +1.96.0 test --manifest-path rust/Cargo.toml --workspace --offline` (445 passed)
- rebuilt the audit parser harness against the patched sources
- 12/12 modern/legacy matrix cells match the oracle
- modern and legacy fork-local same-second probes match the oracle
- existing parent/child same-second coverage remains green

Prepared with AI assistance; I reviewed the implementation and test results.

---
<a href="https://app.blacksmith.sh/ccusage/codesmith/ccusage/pr/1457"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787055794&installation_id=139163553&pr_number=1457&repository=ccusage%2Fccusage&return_to=https%3A%2F%2Fgithub.com%2Fccusage%2Fccusage%2Fpull%2F1457&signature=335fe51931cec5ebaaf14363ebd49e0871ab8180771b8725941a7df40b57257b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
Fixes Codex fork replay detection by using structural boundaries instead of a 1-second heuristic. Prevents double-counting and preserves real fork-local usage while skipping copied history across seconds.

- Bug Fixes
  - Detect replays via adjacent leaf/ancestor `session_meta` or explicit `source.subagent.thread_spawn.parent_thread_id`; set the boundary at a native `task_started` whose `started_at` matches the record second; fall back to the leaf creation second for legacy streams.
  - Skip inherited `token_count` events before the boundary (even when they span multiple seconds) and carry their terminal cumulative totals forward as the live baseline.
  - Keep genuine fork-local same-second `token_count` events; support both modern `last_token_usage` and legacy cumulative-only logs.
  - Preserve established same-second parent/child behavior, including `forked_from_id`, and tolerate malformed `task_started` lines.

<sup>Written for commit ebd45b9f03738eadd3cd1730be6494afeb51e966. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ccusage/ccusage/pull/1457?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Codex replay handling and attribution for forked sessions across modern and legacy log formats.
  - Corrected how token usage totals are determined when events share the same timestamp or span multiple seconds.
  - Enhanced detection of session boundaries and more robust behavior around malformed task markers.

- **Tests**
  - Expanded automated coverage with new Codex fixtures for replay, fork-local same-second cases, and multisecond inherited-prefix scenarios in both supported formats.


ryoppippi:
@pullfrog review it
@codesmith review it
@cubic review it
@coderabbitai review it

ryoppippi:
@pullfrog read all reviews from them and fix them

ryoppippi:
Thanks for digging into this. The underlying bug was real, but this PR has been superseded — the one-second heuristic it replaces no longer exists on `main`.

#1435 and #1438 replaced it with `CodexReplayPlan` in `rust/adapters/codex/src/replay.rs`, which reads the parent session log and subtracts the exact replayed prefix instead of guessing a boundary from a single file. #1428 also moved the adapter to `rust/adapters/codex/`, so the `parser.rs` code this PR rewrites is no longer there in that form.

I checked your `replay-multisecond-modern.jsonl` scenario against current `main` rather than relying on the diff:

| Scenario | Reported total | Correct |
| --- | --- | --- |
| Parent log present in the scanned set | parent 238.7M counted once, child adds 5,500 | yes |
| Parent log absent from the scanned set | 238,705,500 | no — should be 5,500 |

So the common case is already fixed. What remains is narrower than this PR assumes: when the referenced parent log is unavailable, `replay_prefix` returns an empty slice and the parser falls back to the old rewritten-second heuristic, which still misses an inherited prefix spanning multiple seconds. That is a genuine gap and worth closing.

It is better closed inside the `replay.rs` fallback than by reinstating a single-file boundary scan, though. This PR's boundary signal — `task_started.started_at` matching the record second — is itself a heuristic, and a weaker one than reading the parent log, so layering it back over the current design would trade a verified mechanism for an inferred one.

Closing this as superseded. A small PR scoped to just that fallback path would be very welcome, and I would review it happily.

One process note for next time: the verification links point at `docs/reviews/` in your own repository, where the fixtures and the expected values are both authored by the same pass. That is not independent confirmation, and it is what let the multisecond claim stand against a tree that had already moved on. For #1458 I was able to confirm the invariant against 21,815 real `token_count` records, which is the kind of evidence that settles it.

ryoppippi:
Following up: I closed the fallback gap myself in #1501, so you are not left holding it.

The measurements that shaped it, in case they are useful for your accounting work:

- The burst Codex rewrites to the fork instant spans 10 to 40ms across the 9 fork sessions with one in a real log directory, while the child's own first turn follows a pause of 5.8 to 15.3 seconds. Following the run rather than the recorded second separates those two populations by two orders of magnitude, and it is why the multi-second case you found was reachable at all.
- I did not adopt the `task_started.started_at` signal. Of 1,833 real `task_started` records in fork sessions, 135 have `started_at` differing from their record second, and some of those differences are exactly 1 second on native records that merely crossed a tick — so strict equality misclassifies them. That is the kind of thing only real logs surface.
- Before the change, 1 of 210 real fork sessions over-counted by 27,354,644 tokens (2.4x) through the fallback. After it, all 210 agree with exact prefix subtraction.

Thanks for pointing at the right place even though the patch had moved on.

ryoppippi:

PR #1731, opened 2026-09-15
fix(cli): reject a --since date later than --until

### LINKED ISSUE #1729: A --since date later than --until is accepted and prints an empty report
## Summary

A window whose `--since` date is later than its `--until` date is not rejected. The report loads every entry, keeps nothing, and prints an empty table (or `{"daily": [], "totals": {...}}` with `--json`) with exit code 0. That looks like "no usage data" rather than a typo in the flags.

## Reproduce

```sh
ccusage daily --since 2026-09-14 --until 2026-09-01
# empty table, exit 0
ccusage codex monthly --since 20260914 --until 20260901 --json
# {"monthly": [], "totals": {...}}, exit 0
```

Reproduced on main `d341949` and on the published 20.0.20.

## Expected

Reject the window before loading data, the same way #1604 rejects a bound that is not a real date:

```text
The --since date '20260914' is later than the --until date '20260901'.
Run 'ccusage --help' for usage.
```

with exit code 2, for flags, config-file bounds, and a mix of both.

## Notes

`normalize_date_bound` already stores both bounds as `YYYYMMDD`, so the check is a string comparison in `ccusage-cli-parser` after the command is parsed and config is applied. I have this implemented with tests and a docs line and will open the PR right after triage.

ravsau:
Fixes #1729.

A window whose `--since` date is later than its `--until` date previously loaded every entry, kept nothing, and printed an empty report with exit 0, which reads as "no usage data" rather than a typo in the flags.

Both bounds are already normalized to `YYYYMMDD` before storage, so the new `date_window_error` check in `ccusage-cli-parser` is a string comparison. It runs after `last_option_error`, once config bounds and flags are merged, so it covers flags, config-file bounds, and a mix of both, for every report command. The per-command shared-args match that `last_option_error` used is now a shared `report_shared` helper so both checks read the same effective options.

Error and exit code follow the existing date-bound rejection from #1604:

```text
The --since date '20260914' is later than the --until date '20260901'.
Run 'ccusage --help' for usage.
```

Tests cover bare, unified, `session`, `claude session`, `blocks`, and `codex monthly` invocations, an equal-date window, a config `since` with a flag `until`, and a real config file whose reversed window is rejected and then repaired by a flag. The CLI options guide gets one sentence.

Validation: `cargo test -p ccusage-cli-parser` (85 passed), `cargo clippy --all-targets -D warnings`, `cargo fmt --check`, and the release binary end to end.

---
## Summary by cubic
Fixes #1729 by rejecting a date window whose `--since` date is later than its `--until` date. Previously such a window loaded every entry, kept nothing, and printed an empty report with exit 0, which read as missing data rather than a typo.

- The check runs after config bounds and flags are merged, so it covers flags, config-file bounds, and a mix of both for every report command.
- Both bounds are already normalized to `YYYYMMDD`, so the validation is a simple string comparison.
- The error message and non-zero exit code match the existing date-bound rejection.

<sup>Written for commit fb0409fa70edce0b34b7fa7c0a7de7531ffca97b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ccusage/ccusage/pull/1731?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid date ranges are now rejected when the `--since` date is later than the `--until` date.
  - Validation applies consistently across command forms and combinations of command-line and configuration-file options.
  - Equal start and end dates remain accepted.
  - Clear error messages and non-zero exit codes are provided for reversed date ranges.

- **Documentation**
  - Updated CLI option documentation to describe date-range validation behavior.


ryoppippi:
Thank you for the fix. Maintainer review found one `statusline` regression: that command ignores report date bounds, but reversed configured defaults made it fail. I carried your original commit into #1733 with authorship preserved, added the regression fix and test, and ran the full local gates. Closing this copy to avoid duplicate CI and review; #1733 is the merge candidate.

